PR #3, opened 2020-05-27
[ODS-4188] Unable to Reference Post Secondary Institution as Education Organization

jamessmckay:


Application/EdFi.Ods.Security/Authorization/ConcreteAuthorizationContextDataProvider.cs
```diff
@@ -1,4 +1,4 @@
-﻿// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: Apache-2.0
```
blmeyers, line 1: Looks like we added a BOM here?  Files should all be UTF8 only.

blmeyers: Ah ha, there was a BOM, and we removed it, so that's fine.

PR #44, opened 2020-06-26
[ODS-4046] Recreating student, staff or parent after deletion returns 400 Bad Request

gmcelhanon:
This PR implements configuration file-based cache expiration settings for the Descriptors cache and Person UniqueId/USI cache.

See also https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pull/27

### DIFF
--- a/Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
+++ b/Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
@@ -2,7 +2,8 @@
 // Licensed to the Ed-Fi Alliance under one or more agreements.
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
- 
+
+using System;
 using System.Linq;
 using System.Reflection;
 using System.Web.Http;
@@ -47,14 +48,19 @@ namespace EdFi.Ods.Api.Startup.Features.Installers
 {
     public class CoreApiInstaller : IWindsorInstaller
     {
+        private const string PersonCacheAbsoluteExpirationSecondsKey = "caching:personUniqueIdToUsi:absoluteExpirationSeconds";
+        private const string PersonCacheSlidingExpirationSecondsKey = "caching:personUniqueIdToUsi:slidingExpirationSeconds";
+        
+        private readonly IConfigValueProvider _configValueProvider;
         private readonly Assembly _apiAssembly;
         private readonly IApiConfigurationProvider _apiConfigurationProvider;
         private readonly Assembly _standardAssembly;
 
-        public CoreApiInstaller(IAssembliesProvider assembliesProvider, IApiConfigurationProvider apiConfigurationProvider)
+        public CoreApiInstaller(IAssembliesProvider assembliesProvider, IApiConfigurationProvider apiConfigurationProvider, IConfigValueProvider configValueProvider)
         {
             Preconditions.ThrowIfNull(assembliesProvider, nameof(assembliesProvider));
             _apiConfigurationProvider = Preconditions.ThrowIfNull(apiConfigurationProvider, nameof(apiConfigurationProvider));
+            _configValueProvider = Preconditions.ThrowIfNull(configValueProvider, nameof(configValueProvider));
 
             var installedAssemblies = assembliesProvider.GetAssemblies().ToList();
 
@@ -235,11 +241,13 @@ namespace EdFi.Ods.Api.Startup.Features.Installers
 
         private void RegisterPersonUniqueIdToUsiCache(IWindsorContainer container)
         {
-            container.Register(
-                Component
-                    .For<IPersonUniqueIdToUsiCache>()
-                    .ImplementedBy<PersonUniqueIdToUsiCache>()
-                    .DependsOn(Dependency.OnValue("synchronousInitialization", false)));
+            container.Register(Component.For<IPersonUniqueIdToUsiCache>()
+                .ImplementedBy<PersonUniqueIdToUsiCache>()
+                .DependsOn(Dependency.OnValue("slidingExpiration",
+                    TimeSpan.FromSeconds(Convert.ToInt32(_configValueProvider.GetValue(PersonCacheSlidingExpirationSecondsKey) ?? "14400"))))
+                .DependsOn(Dependency.OnValue("absoluteExpirationPeriod", 
+                    TimeSpan.FromSeconds(Convert.ToInt32(_configValueProvider.GetValue(PersonCacheAbsoluteExpirationSecondsKey) ?? "86400"))))
+                .DependsOn(Dependency.OnValue("synchronousInitialization", false))); 
         }
 
         private void RegisterDescriptorCache(IWindsorContainer container)
--- a/Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
+++ b/Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
@@ -48,6 +48,7 @@ namespace EdFi.Ods.Api.Startup.Features.Installers
 {
     public class CoreApiInstaller : IWindsorInstaller
     {
+        private const string DescriptorCacheAbsoluteExpirationSecondsKey = "caching:descriptors:absoluteExpirationSeconds";
         private const string PersonCacheAbsoluteExpirationSecondsKey = "caching:personUniqueIdToUsi:absoluteExpirationSeconds";
         private const string PersonCacheSlidingExpirationSecondsKey = "caching:personUniqueIdToUsi:slidingExpirationSeconds";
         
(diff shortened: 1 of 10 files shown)


Application/EdFi.Ods.Api/Caching/PersonUniqueIdToUsiCache.cs
```diff
@@ -290,7 +316,7 @@ private async Task InitializePersonTypeValueMapsAsync(IdentityValueMaps entry, s
                 string cacheKey = GetPersonTypeIdentifiersCacheKey(personType, context);
 
                 //Now that it's loaded extend the cache expiration.
-                _cacheProvider.Insert(cacheKey, entry, DateTime.MaxValue, TimeSpan.FromHours(4));
+                _cacheProvider.Insert(cacheKey, entry, GetAbsoluteExpiration(), _slidingExpiration);
```
blmeyers, line 319: The original change that spurred all these cache issues is the switch to use ConcurrentDictionaryCacheProvider, which doesn't have any sort of cache expiration.  In this case, the parameters you're updating here are completely ignored in that implementation.
```csharp
public void Insert(string key, object value, DateTime absoluteExpiration, TimeSpan slidingExpiration)
{
    _cacheDictionary[key] = value;
}
```
The underlying cache provider would need to be updated to actually use these parameters.

Application/EdFi.Ods.Api/Caching/DescriptorsCache.cs
```diff
@@ -55,7 +57,7 @@ private bool HasLastDatabaseSynchronizationExpired
                 }
 
                 var totalSeconds = (currentTime - lastSynced).TotalSeconds;
-                return totalSeconds > DatabaseSynchronizationExpirationSeconds;
+                return totalSeconds > _absoluteExpirationPeriod.TotalSeconds;
```
blmeyers, line 60: This property is probably not well named, its really used more like a quiet period setting.  Requests for descriptors from the cache will refresh the descriptors if there's a miss that gets resolved, and will then attempt to refresh the entire cache under the assumption that the cache is out of sync.  If this quiet period is still in effect, it will not do that entire cache refresh.  This doesn't actually cause the cache to get cleared out, we'd have to update the places where we're inserting to the cache as well as updating the underlying cache provider to respect these parameters.

Application/EdFi.Ods.Api/Caching/PersonUniqueIdToUsiCache.cs
gmcelhanon: I looked at doing this, but it no longer fit into what I considered to be a simple approach, which is what we were shooting for here.

Adding a behavior that clears the entire `ConcurrentDictionary` after a particular period of time seems a bit dangerous. The cache providers are necessarily registered with the _singleton_ lifestyle so that the same instance is made available everywhere it is injected. To add a behavior that periodically clears the entire cache (for potentially many use cases, depending on the actual IoC configuration) seems like an unexpected behavior that could inadvertently cause someone trouble. As a more sensible alternative, implementing the expiration features so that it supports expiration of each cache entry certainly very quickly goes beyond the "simple" approach as well.

Another option, would be to create a copy (or override) of the `ConcurrentDictionaryCacheProvider` named `ExpiringConcurrentDictionaryCacheProvider`, and implement a simple approach that clears all the cache entries periodically. Then we'd modify the `DescriptorsCache` registration to use that version explicitly. I think it would also make sense to register it with a _transient_ lifestyle.

Another alternative would be to re-engineer the Descriptors cache to work more like the Person cache does -- to use the `AspNetCacheProvider` for a main entry for each ODS context, and then use `ConcurrentDictionary` properties on the cache entry for managing the Descriptor values. But again, this seems likely out of scope for the ticket.

A final alternative is to just not do anything right now and to put these two tickets back into the backlog and revisit after the release.

@vimayya Thoughts as to which direction you'd like this to go?

Application/EdFi.Ods.Api/Caching/DescriptorsCache.cs
gmcelhanon: So, this is just primarily to stop bad client data from forcing full descriptor cache refreshes with every miss. The full refresh will only happen once a minute, at most (as originally written).

blmeyers: Yep, exactly.

Application/EdFi.Ods.Api/Caching/PersonUniqueIdToUsiCache.cs
vimayya: @gmcelhanon I think ExpiringConcurrentDictionaryCacheProvider makes sense. Currently in the ConcurrentDictionaryCacheProvider's implementation of Insert method, the expiry isn't taking place which is what I wanted to address in the release. If we redesign the way we keep cache up to date later, switching back to ConcurrentDictionaryCacheProvider is easy enough.

Application/EdFi.Ods.Api/Caching/PersonUniqueIdToUsiCache.cs
```diff
@@ -300,6 +326,10 @@ private async Task InitializePersonTypeValueMapsAsync(IdentityValueMaps entry, s
             }
         }
 
+        private DateTime GetAbsoluteExpiration() => _absoluteExpirationPeriod == TimeSpan.Zero
+            ? DateTime.MaxValue 
+            : DateTime.Now.Add(_absoluteExpirationPeriod);
```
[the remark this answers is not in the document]
gmcelhanon: Changed.

Application/EdFi.Ods.Api/Caching/DescriptorsCache.cs
gmcelhanon: Reverted all changes to DescriptorsCache -- and implemented `ExpiringConcurrentDictionaryCacheProvider`.

Application/EdFi.Ods.Api/Caching/ExpiringConcurrentDictionaryCacheProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+        private Timer _timer;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExpiringConcurrentDictionaryCacheProvider" /> class using the
+        /// specified recurring expiration period.
+        /// </summary>
+        /// <param name="expirationPeriod">The recurring expiration period for all of the entries in the cache.</param>
+        public ExpiringConcurrentDictionaryCacheProvider(TimeSpan expirationPeriod)
+        {
+            _timer = new Timer(CacheExpired, null, expirationPeriod, expirationPeriod);
```
blmeyers, line 29: This class is explicitly instantiated as a dependency of a singleton, so you should be safe with the timer consistently remaining static.  I really dislike the use of a timer here and this seems to be yet another unintended side effect of rolling our own caching.  Given this is caching related and any of the usual issues with timers should also result in the cache being dropped this is probably ok as a short term solution until we determine a more consistent way to handle the caching.

Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
```diff
@@ -235,22 +242,26 @@ private void RegisterUniqueIdToUsiValueMapper(IWindsorContainer container)
 
         private void RegisterPersonUniqueIdToUsiCache(IWindsorContainer container)
         {
-            container.Register(
-                Component
-                    .For<IPersonUniqueIdToUsiCache>()
-                    .ImplementedBy<PersonUniqueIdToUsiCache>()
-                    .DependsOn(Dependency.OnValue("synchronousInitialization", false)));
+            container.Register(Component.For<IPersonUniqueIdToUsiCache>()
```
blmeyers, line 245: The original ticket was specifically about people not being able to recreate after delete, so shouldn't this cache also be consuming the ExpiringConcurrentDictionaryCacheProvider?

Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
```diff
@@ -235,22 +242,26 @@ private void RegisterUniqueIdToUsiValueMapper(IWindsorContainer container)
...
+                .ImplementedBy<PersonUniqueIdToUsiCache>()
+                .DependsOn(Dependency.OnValue("slidingExpiration",
```
blmeyers, line 247: Since we can't actually do sliding expiration with our current cache implementations, this should be removed everywhere.

Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
blmeyers: As per standup, this is still using the Asp.Net cache, so it correctly handles expirations.

Application/EdFi.Ods.Api.Startup/Features/Installers/CoreApiInstaller.cs
blmeyers: As per standup, this is still using the Asp.Net cache, so it correctly handles expirations.

blmeyers:
The one broken build is due to an intermittent issue resolved with https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/52, which is already merged in now.

PR #46, opened 2020-06-29
[ODS-3164] 3.1 DisciplineIncident: cannot change staffReference

tnunnerydl:
… value.

### DIFF
--- a/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/Entities_PropertyAccessors.mustache
+++ b/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/Entities_PropertyAccessors.mustache
@@ -68,6 +68,11 @@
             }
             set
             {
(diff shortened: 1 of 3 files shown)


Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/Entities_PropertyAccessors.mustache
```diff
@@ -68,6 +68,11 @@
             }
             set
             {
+                {{#UsiProperty}}
```
blmeyers, line 71: This entire template section is in an #UsiProperty check, so no need to verify it again here.  Just add the new if statement.

tnunnerydl: Done

PR #50, opened 2020-07-01
[ODS-4343] Update the default samples directory

stephenfuqua:


Utilities/DataLoading/EdFi.BulkLoadClient.Console/Properties/Settings.Designer.cs
```diff
@@ -157,7 +157,7 @@ internal sealed partial class Settings : global::System.Configuration.Applicatio
         
         [global::System.Configuration.UserScopedSettingAttribute()]
         [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
-        [global::System.Configuration.DefaultSettingValueAttribute("..\\..\\..\\..\\..\\..\\Ed-Fi-Standard\\v3.2\\Samples\\Sample XML")]
+        [global::System.Configuration.DefaultSettingValueAttribute("..\\..\\..\\..\\..\\..\\Ed-Fi-Standard\\Samples\\Sample XML")]
```
blmeyers, line 160: This is a generated file.  To fix these the file Utilities/DataLoading/EdFi.BulkLoadClient.Console/Properties/Settings.settings needs to be updated, which will then regenerate this file when you right click "Run Custom Tool".  I found this option to run custom tool is only there for VS2017 and not with VS2019, which makese some sense since using settings is sort of an older concept.  Likely should be another ticket opened to remove all usage of settings, and stick to just config and cli parameters.

You could probably just manually fix the Settings.settings and this designer file to get around the 2019 limitation.1

Note, there are also two other places in this Settings.settings file that has paths to v3.2 that should also be fixed.

PR #53, opened 2020-07-06
[ODS-4316] [ODS-4301] Authorization views for CommunityProvider and CommunityOrganization do not follow required naming conventions

jamessmckay:


Application/EdFi.Ods.Security/AuthorizationStrategies/Relationships/Filters/RelationshipsAuthorizationFilters.cs
```diff
@@ -47,7 +47,7 @@ private static FilterApplicationDetails CreateClaimValuePropertyFilter(string pr
                         "LocalEducationAgencyIdToStudentUSI",
                         @"StudentUSI IN (
                             SELECT {newAlias1}.StudentUSI 
-                            FROM [auth].[LocalEducationAgencyIdToStudentUSI] {newAlias1} 
+                            FROM auth.LocalEducationAgencyIdToStudentUSI {newAlias1}
```
blmeyers, line 50: With the refactors away from bulk, this actually isn't sql anymore, it appears to just be some form of generic query going into a FilterDefinition.  Probably worth another ticket to figure out if we need both variations of the filter, or if we could simplify down to just one with Bulk gone.

PR #63, opened 2020-07-10
[ODS-4344] Add GrandBend namespace to sandbox client

jamessmckay:


Application/EdFi.Ods.Admin/Initialization/NamespacePrefixInitializationModel.cs
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.Ods.Admin.Initialization
+{
+    public class NamespacePrefixInitializationModel
+    {
+        public string Name { get; set; }
```
blmeyers, line 10: This should probably be named "NamespacePrefix" to be clear what it represents, we're not naming the prefixes.

Application/EdFi.Ods.Sandbox/Repositories/ClientAppRepo.cs
```diff
@@ -463,6 +481,7 @@ public Vendor CreateOrGetVendor(string userEmail, string userName, IEnumerable<s
                         }
                     }
 
+                    await context.SaveChangesAsync();
```
blmeyers, line 484: This method didn't used to save, but it probably makes sense for it to ensure it saves after adding a new Vendor.  Shouldn't this Save call be inside the if check for "vendor == null" though?  No need to save if we're just retrieving an existing vendor.

Application/EdFi.Ods.Sandbox/Repositories/ClientAppRepo.cs
```diff
@@ -436,14 +437,31 @@ public void SetDefaultVendorOnUserFromEmailAndName(string userEmail, string user
...
+                var user = await context.Users.SingleAsync(u => u.Email.Equals(userEmail));
+
+                user.Vendor = vendor;
+                await context.SaveChangesAsync();
+            }
+        }
+
+        public Vendor CreateOrGetVendor(string userEmail, string userName, IEnumerable<string> namespacePrefixes)
+        {
+            return CreateOrGetVendorAsync(userEmail, userName, namespacePrefixes).GetResultSafely();
```
blmeyers, line 454: Is this the right time to start using async?  Shouldn't we introduce async to this project in a separate ticket rather than mixing usages?

Application/EdFi.Ods.Admin/Initialization/InitializationEngine.cs
```diff
@@ -79,23 +80,23 @@ public void CreateUsers()
             }
         }
 
-        private void CreateUser(UserInitializationModel user)
+        private async void CreateUser(UserInitializationModel user)
```
blmeyers, line 83: We shouldn't be using async void, this should return a Task, and then resharper is likely to tell you to just return the Task directly from SetDefaultVendorOnUserFromEmailAndNameAsync instead of awaiting it.  I'm guessing you were trying to avoid async from leaking out, but this is another good reason to handle the conversion to use async separately.

Application/EdFi.Ods.Admin/Initialization/InitializationEngine.cs
```diff
@@ -79,23 +80,23 @@ public void CreateUsers()
...
             {
                 Roles.AddUserToRole(user.Email, role);
             }
 
             WebSecurityService.UpdatePasswordAndActivate(user.UserName, user.Password);
-            _clientAppRepo.SetDefaultVendorOnUserFromEmailAndName(user.Email, user.Name);
+
+            var namespacePrefixes = user.NamespacePrefixes.Select(nsp => nsp.Name).ToList();
+
+            await _clientAppRepo.SetDefaultVendorOnUserFromEmailAndNameAsync(user.Email, user.UserName, namespacePrefixes);
```
blmeyers, line 99: If we do use async, we need to be configure awaiting false everywhere, this is a library so we don't want to mess with SynchronizationContext.

Application/EdFi.Ods.Sandbox/Repositories/ClientAppRepo.cs
jamessmckay: updated

Application/EdFi.Ods.Admin/Initialization/NamespacePrefixInitializationModel.cs
jamessmckay: done

Application/EdFi.Ods.Sandbox/Repositories/ClientAppRepo.cs
jamessmckay: done

Application/EdFi.Ods.Admin/Initialization/InitializationEngine.cs
jamessmckay: done

Application/EdFi.Ods.Admin/Initialization/InitializationEngine.cs
jamessmckay: done

PR #66, opened 2020-07-16
[ODS-4365] Bring in Certification namespace update 

tnunnerydl:


blmeyers:
This was missed when merging in ODS-4365, and includes approval test updates for local only tests.  Merging in to ensure these tests stay correct, but will have no impact on previously merged changes to TPDM.

PR #150, opened 2020-09-08
[ODS-4506][ODS-4496] Fix Change Queries Postman Tests failures

blmeyers:


Application/EdFi.Ods.ChangeQueries/Controllers/AvailableChangeVersionsController.cs
```diff
@@ -40,7 +43,13 @@ public IActionResult Get()
                 return NotFound();
             }
 
-            return Ok(_availableChangeVersionProvider.GetAvailableChangeVersion());
+            // Explicitly serialize the response to remain backwards compatible with pre .net core
+            return new ContentResult
+            {
+                Content = JsonConvert.SerializeObject(_availableChangeVersionProvider.GetAvailableChangeVersion()),
+                ContentType = MediaTypeNames.Application.Json,
+                StatusCode = (int)HttpStatusCode.OK
```
jamessmckay, line 51: HttpStatusCode is actually net netframework enum, the one we want to use is StatusCodes. For example StatusCodes.Status200Ok

Application/EdFi.Ods.ChangeQueries/Controllers/DeletesController.cs
```diff
@@ -43,7 +46,13 @@ public IActionResult Get(string schema, string resource, [FromQuery] UrlQueryPar
 
             var result = _getDeletedResourceIdsRepository.Execute(schema, resource, queryParameter);
 
-            return Ok(result);
+            // Explicitly serialize the response to remain backwards compatible with pre .net core
+            return new ContentResult
+            {
+                Content = JsonConvert.SerializeObject(result),
+                ContentType = MediaTypeNames.Application.Json,
+                StatusCode = (int)HttpStatusCode.OK
```
jamessmckay, line 54: see above

jamessmckay (changes requested):

Application/EdFi.Ods.ChangeQueries/Controllers/AvailableChangeVersionsController.cs
blmeyers: StatusCodes is fine for here since this is already ASP.Net Core code, but bear in mind that HttpStatusCode is still available in .Net Core (and the upcoming .Net 5.0), and is what you should still be using if you're not pulling in the ASP.Net Core framework.

jamessmckay: Interesting, however in our case we reference the netcore framework.

jamessmckay (approved):

PR #154, opened 2020-09-10
[ODS-4295] Convert Profiles Feature to NET Core

jamessmckay:


Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
```diff
@@ -213,6 +221,7 @@ public virtual async Task<IActionResult> Get(Guid id)
         [ProducesResponseType(StatusCodes.Status204NoContent)]
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
         [ProducesResponseType(StatusCodes.Status412PreconditionFailed)]
+        [Produces(MediaTypeNames.Application.Json)]
```
tnunnerydl, line 224: Is there a way to set this globally, since the majority, if not all responses are return in json?

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
```diff
@@ -239,6 +248,7 @@ public virtual async Task<IActionResult> Put([FromBody] TPutRequest request, Gui
             }
 
             Response.Headers[HeaderConstants.ETag] = GetEtag(result.ETag).ToString();
+            Response.GetTypedHeaders().ContentType = new MediaTypeHeaderValue(MediaTypeNames.Application.Json);
```
tnunnerydl, line 251: Why do we need to set ContentType and decorate with the Produce attribute?  Aren't they doing the same thing?

Application/EdFi.Ods.Features/Container/Modules/CompositesModule.cs
```diff
@@ -10,11 +10,11 @@
 using EdFi.Ods.Common.Constants;
 using EdFi.Ods.Common.Container;
 using EdFi.Ods.Common.Metadata;
+using EdFi.Ods.Common.Models;
```
tnunnerydl, line 13: It doesn't appear that the namespace is used here.  Needs to be removed.

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
jamessmckay: yes however in this case, change only get and get all depending on the profile if applied. Otherwise to set globally we attach the attribute to the controller

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
jamessmckay: profiles has a differnt content type, and the decoration breaks post and put

jamessmckay: ohh i see what you saying i will remove this line on put and post since we are decorating the method

tnunnerydl (approved):

PR #156, opened 2020-09-11
[ODS-4505] Fix config transformation to address postgres build

tbmeador-dlp:


jamessmckay (approved):

Application/EdFi.Common/Configuration/DatabaseEngine.cs
```diff
@@ -21,6 +21,17 @@ public DatabaseEngine(string value, string displayName, string scriptsFolderName
         }
 
         public string ScriptsFolderName { get; private set; }
+
+        public static DatabaseEngine TryParseEngine(string value)
```
blmeyers, line 25: If we want an exception why not just use Parse?  Does it have to be this message and a NotSupportedException?

Application/EdFi.Ods.Common/Configuration/ApiSettings.cs
```diff
@@ -18,7 +18,7 @@ public class ApiSettings
 
         public ApiSettings()
         {
-            _databaseEngine = new Lazy<DatabaseEngine>(() => DatabaseEngine.FromValue(Engine));
+            _databaseEngine = new Lazy<DatabaseEngine>(() => DatabaseEngine.TryParseEngine(Engine));
```
blmeyers, line 21: Should this be outside of a Lazy so it fails immediately for bad configuration?

Application/EdFi.Security.DataAccess/Contexts/PostgresSecurityContext.cs
```diff
@@ -9,9 +9,22 @@
 using System.Data.Entity.ModelConfiguration.Conventions;
 using EdFi.Ods.Common;
 using EdFi.Ods.Common.Utils.Extensions;
+using Npgsql;
 
 namespace EdFi.Security.DataAccess.Contexts
 {
+    class NpgsqlConfiguration : DbConfiguration
```
blmeyers, line 16: See previous comment about making this in EdFi.Common.

blmeyers (changes requested):

Application/EdFi.Common/Configuration/DatabaseEngine.cs
tbmeador-dlp: Here I was just following the same pattern as the `ApiMode`

blmeyers (approved):

PR #163, opened 2020-09-16
[ODS-4332] InstanceYear Api Support

frenchyjef:
Adding support for "InstanceYearSpecific" startup strategy.

Note that the OpenApi routing for "InstanceYearSpecific" is not implemented. A second pull request will be submitted in October once the OpenApi routing for "YearSpecific" and "DistrictSpecific" has been implemented by Ed-Fi team.

### DIFF
--- a/Application/EdFi.Ods.Api/Container/Modules/InstanceYearSpecificModule.cs
+++ b/Application/EdFi.Ods.Api/Container/Modules/InstanceYearSpecificModule.cs
@@ -1,0 +1,62 @@
+﻿// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using Autofac;
+using Autofac.Core;
+using EdFi.Common.Configuration;
+using EdFi.Ods.Api.Caching;
+using EdFi.Ods.Api.Conventions;
+using EdFi.Ods.Api.Filters;
+using EdFi.Ods.Common.Caching;
+using EdFi.Ods.Common.Configuration;
+using EdFi.Ods.Common.Container;
+using EdFi.Ods.Common.Context;
+using EdFi.Ods.Common.Database;
+using EdFi.Security.DataAccess.Caching;
+using EdFi.Security.DataAccess.Repositories;
+using Microsoft.AspNetCore.Mvc.ApplicationModels;
+using Microsoft.AspNetCore.Mvc.Filters;
+
+namespace EdFi.Ods.Api.Container.Modules
+{
+    public class InstanceYearSpecificModule : ConditionalModule
+    {
+        public InstanceYearSpecificModule(ApiSettings apiSettings)
+            : base(apiSettings, nameof(InstanceYearSpecificModule)) { }
+
+        public override bool IsSelected() => ApiSettings.GetApiMode() == ApiMode.InstanceYearSpecific;
+
+        public override void ApplyConfigurationSpecificRegistrations(ContainerBuilder builder)
+        {
+            builder.RegisterType<TokenControllerRouteConvention>()
+                .As<IApplicationModelConvention>()
+                .SingleInstance();
+
+            builder.RegisterType<InstanceIdContextFilter>()
+                .As<IFilterMetadata>();
+
+            builder.RegisterType<InstanceIdContextProvider>()
+                .As<IInstanceIdContextProvider>()
+                .As<IHttpContextStorageTransferKeys>();
+
+            builder.RegisterType<InstanceYearSpecificAdminDatabaseNameReplacementTokenProvider>()
+                .As<IAdminDatabaseNameReplacementTokenProvider>();
+
+            builder.RegisterType<InstanceYearSpecificSecurityDatabaseNameReplacementTokenProvider>()
+                .As<ISecurityDatabaseNameReplacementTokenProvider>();
+
+            builder.RegisterType<InstanceYearSpecificDatabaseNameReplacementTokenProvider>()
+                .As<IDatabaseNameReplacementTokenProvider>();
+
+            builder.RegisterType<InstanceSecurityRepository>()
+                .As<ISecurityRepository>()
+                .SingleInstance();
+
+            builder.RegisterType<InstanceSecurityRepositoryCache>()
+                .As<IInstanceSecurityRepositoryCache>()
+                .SingleInstance();
+        }
+    }
+}
--- a/Application/EdFi.Ods.Api/Container/Modules/PersistenceModule.cs
+++ b/Application/EdFi.Ods.Api/Container/Modules/PersistenceModule.cs
@@ -31,10 +31,6 @@ namespace EdFi.Ods.Api.Container.Modules
     {
         protected override void Load(ContainerBuilder builder)
         {
-            builder.RegisterType<AdminDatabaseConnectionStringProvider>()
-                .As<IAdminDatabaseConnectionStringProvider>()
-                .SingleInstance();
-
             builder.Register(c => new MemoryCache(new MemoryCacheOptions()))
                 .As<IMemoryCache>()
                 .SingleInstance();
@@ -124,10 +120,6 @@ namespace EdFi.Ods.Api.Container.Modules
 
             builder.RegisterGeneric(typeof(UpsertEntity<>))
                 .As(typeof(IUpsertEntity<>))
-                .SingleInstance();
-
-            builder.RegisterType<SecurityDatabaseConnectionStringProvider>()
-                .As<ISecurityDatabaseConnectionStringProvider>()
                 .SingleInstance();
 
             builder.RegisterType<UniqueIdToUsiValueMapper>().As<IUniqueIdToUsiValueMapper>()
--- a/Application/EdFi.Ods.Api/Middleware/EdFiOAuthAuthenticationHandler.cs
+++ b/Application/EdFi.Ods.Api/Middleware/EdFiOAuthAuthenticationHandler.cs
@@ -6,8 +6,10 @@
 using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using EdFi.Ods.Api.Providers;
+using EdFi.Ods.Common.Context;
 using EdFi.Ods.Common.Security;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.Extensions.Logging;
@@ -19,6 +21,7 @@ namespace EdFi.Ods.Api.Middleware
     {
         private readonly IApiKeyContextProvider _apiKeyContextProvider;
         private readonly IAuthenticationProvider _authenticationProvider;
+        private readonly IInstanceIdContextProvider _instanceIdContextProvider;
 
         public EdFiOAuthAuthenticationHandler(
             IOptionsMonitor<AuthenticationSchemeOptions> options,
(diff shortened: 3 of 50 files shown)


Application/EdFi.Admin.DataAccess/Providers/AdminDatabaseConnectionStringProvider.cs
```diff
@@ -4,20 +4,42 @@
...
+        {
+            if (ApiSettings.GetApiMode() == ApiMode.InstanceYearSpecific)
+            {
+                string dbName = ConnectionStringName;
+                string instanceId = _instanceIdContextProvider.GetInstanceId();
+                if (string.IsNullOrEmpty(instanceId))
+                {
+                    throw new InvalidOperationException("Expected instanceId is null or empty.");
+                }
+                string connStr = _configConnectionStringsProvider.GetConnectionString(ConnectionStringName).Replace(dbName, $"{dbName}_{instanceId}");
```
blmeyers, line 37: This is assuming the Database name always has to be "EdFi_Admin".  This would probably be better to match the pattern of ODS, and use a replacement token approach so the implementation can prefix their databases however they want.

Application/EdFi.Security.DataAccess/Providers/SecurityDatabaseConnectionStringProvider.cs
```diff
@@ -4,19 +4,41 @@
...
+        {
+            if (ApiSettings.GetApiMode() == ApiMode.InstanceYearSpecific)
+            {
+                string dbName = ConnectionStringName;
+                string instanceId = _instanceIdContextProvider.GetInstanceId();
+                if (string.IsNullOrEmpty(instanceId))
+                {
+                    throw new InvalidOperationException("Expected instanceId is null or empty.");
+                }
+                string connStr = _configConnectionStringsProvider.GetConnectionString(ConnectionStringName).Replace(dbName, $"{dbName}_{instanceId}");
```
blmeyers, line 36: Similar to Admin, this should probably use a replacement token approach.

Application/EdFi.Security.DataAccess/Providers/SecurityDatabaseConnectionStringProvider.cs
```diff
@@ -4,19 +4,41 @@
...
 #if NETSTANDARD
+using System;
 using EdFi.Ods.Common.Configuration;
+using EdFi.Ods.Common.Context;
 
 namespace EdFi.Security.DataAccess.Providers {
     public class SecurityDatabaseConnectionStringProvider: ISecurityDatabaseConnectionStringProvider
     {
+        private const string ConnectionStringName = "EdFi_Security";
+        protected readonly ApiSettings ApiSettings;
```
blmeyers, line 15: Should be private visibility, and adjust the name to match.

Application/EdFi.Admin.DataAccess/Providers/AdminDatabaseConnectionStringProvider.cs
```diff
@@ -4,20 +4,42 @@
 // See the LICENSE and NOTICES files in the project root for more information.
 
 #if NETSTANDARD
+using System;
 using EdFi.Ods.Common.Configuration;
+using EdFi.Ods.Common.Context;
 
 namespace EdFi.Admin.DataAccess.Providers
 {
     public class AdminDatabaseConnectionStringProvider : IAdminDatabaseConnectionStringProvider
```
blmeyers, line 13: This project generates a nuget package, and is also consumed by other tools such as Admin App. Adding these Ods specific dependencies will cause some trouble for that use case. Most likely Ods would need to create its own new version of IAdminDatabaseConnectionStringProvider implemented with this new support, and then register that for ODS/API, leaving the original unmodified.

blmeyers (changes requested):
In general the implementation looks pretty good, aligns with the patterns very well.  Other than the Security caching concern raised in a comment, I have a few higher level concerns I'd like to raise:

1. This approach duplicates Admin and Security databases, which hasn't previously been done before.  If you run with this api mode, effectively the only thing you're sharing resources on is the Web Api itself.  What is the benefit of this approach versus just deploying another web api to go along with all the databases, and keeping the entire system stand alone?  Would the current community container work for the ODS/API be another potential solution to solve this problem in an efficient manner?
2. Lack of test coverage for any of the new components.  It looks like the only test changes were to keep the existing tests running and passing.

Also @jamessmckay, notice how many times a new api mode had to replicate the route modifications.  Something still isn't quite right with that design, this seems far too repetitive to make global route changes.

frenchyjef:
> 1. This approach duplicates Admin and Security databases, which hasn't previously been done before.  If you run with this api mode, effectively the only thing you're sharing resources on is the Web Api itself.  What is the benefit of this approach versus just deploying another web api to go along with all the databases, and keeping the entire system stand alone?  Would the current community container work for the ODS/API be another potential solution to solve this problem in an efficient manner?

From a SaaS provider perspective, we must provide high availability, performance at scale and data isolation (security) while balancing the complexity of operating a multi-tenant environment.  If we had to create two instances (minimum, more for larger tenant) of a Web API per tenant, then we would essentially have a minimum of 100 instances of the Web Api for 50 tenants which requires more VM to run on.  To make our SaaS environment more efficient and simple to operate, we instantiate a minimum of 5 instances of the Web API that all 50 tenants share and then we scale up and down the number of instances based on the current load (a tenant might require more performance than other at different time during the day).  Also, since we run multiple version of the Web Api side-by-side (v2.6, v3.4, v5.0, etc.), you can see how this could become more complex/difficult to manage then having a single set of instances per Web API version shared by their respective tenants.

Thoughts?

vimayya:
@frenchyjef Are you hosting all your client databases on one database server ? without having a way to target multiple servers via one api setting (doesn't look like it was included in this PR), you might still need multiple api server configurations. Is that something that is coming up in another PR?

frenchyjef:
This PR targets a single SQL Server and the InstanceId is appended to EdFi_Admin_InstanceID, EdFi_Security_InstanceID and EdFi_Ods_InstanceID_Year.

We have implemented a beta version of a "dynamic connection string" connection string which would use an Azure KeyVault "secrets" to store each connection string where each secret would have the InstanceId appended. For example, EdFi_Admin_InstanceID, EdFi_Security_InstanceID and EdFi_Ods_InstanceID.  If someone wants to use AWS, or some other secret store, they can add their own implementation.

This "dynamic connection string" feature was implemented in v3.4 and will need to be re-implemented for v5.1 to align with .NET Core. However, I would like to discuss this feature with you before re-implementing in v5.1 to make sure that it is feature that is considered by Ed-Fi.

frenchyjef:
@blmeyers @vimayya We have implemented the changes you requested. Please let us know if you have any feedback before merging into your main branch.

blmeyers (changes requested):
Sorry for the delay on getting back to this.  Most of the feedback is minor, just adjusting for coding standards.  I think your changes have uncovered some gaps in extensibility, particularly around the Security Repository architecture, so I'll be spinning a few tickets off to investigate some of that further.  The approach you took works for what you're trying to do, but that you had to go to some of the lengths you did and duplicate some of that repository code is a good sign that the current structure needs some attention.

You'll need to catch this branch up, as well as please follow the directions at https://techdocs.ed-fi.org/display/ETKB/Signing+Git+Commits to make sure you've signed the CLA and are signing your commits for verification.

Application/EdFi.Ods.Api/Container/Modules/InstanceYearSpecificModule.cs
```diff
@@ -0,0 +1,67 @@
...
+                .As<IDatabaseNameReplacementTokenProvider>();
+
+            builder.RegisterType<InstanceSecurityRepository>()
+                .As<ISecurityRepository>();
+
+            builder.RegisterType<InstanceSecurityRepositoryCache>()
+                .WithParameter(
+                    new ResolvedParameter(
+                        (p, c) => p.ParameterType == typeof(IDescriptorsCache),
+                        (p, c) => c.Resolve<ConcurrentDictionaryCacheProvider>()))
```
blmeyers, line 61: Copy and paste from the descriptor registration?  There are no parameters of type "IDescriptorsCache" on this class, did you mean to do this for ICacheProvider?

Application/EdFi.Ods.Api/Container/Modules/PersistenceModule.cs
```diff
@@ -31,9 +31,10 @@ public class PersistenceModule : Module
     {
         protected override void Load(ContainerBuilder builder)
         {
-            builder.RegisterType<AdminDatabaseConnectionStringProvider>()
-                .As<IAdminDatabaseConnectionStringProvider>()
-                .SingleInstance();
+            //NOTE: Need to make is conditional
+            //builder.RegisterType<AdminDatabaseConnectionStringProvider>()
+            //    .As<IAdminDatabaseConnectionStringProvider>()
+            //    .SingleInstance();
```
blmeyers, line 38: Remove these lines now that they're handled in your main module.

Application/EdFi.Ods.Api/Container/Modules/PersistenceModule.cs
```diff
@@ -106,8 +107,9 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterGeneric(typeof(UpsertEntity<>))
                 .As(typeof(IUpsertEntity<>));
 
-            builder.RegisterType<SecurityDatabaseConnectionStringProvider>()
-                .As<ISecurityDatabaseConnectionStringProvider>();
+            //NOTE: Need to make is conditional
+            //builder.RegisterType<SecurityDatabaseConnectionStringProvider>()
+            //    .As<ISecurityDatabaseConnectionStringProvider>();
```
blmeyers, line 113: Remove these lines now that they're handled in your main module.

Application/EdFi.Ods.Security/Container/Modules/SecurityPersistenceModule.cs
```diff
@@ -74,8 +74,9 @@ protected override void Load(ContainerBuilder builder)
                 .As<IEducationOrganizationIdentifiersValueMapper>()
                 .AsSelf();
 
-            builder.RegisterType<SecurityRepository>()
-                .As<ISecurityRepository>();
+            //NOTE: Need to be make is conditional
+            //builder.RegisterType<SecurityRepository>()
+            //    .As<ISecurityRepository>();
```
blmeyers, line 80: Remove these lines now that they're handled in your main module.

Application/EdFi.Ods.Tests/EdFi.Ods.Api/Filters/InstanceIdContextFilterTests.cs
```diff
@@ -146,6 +146,8 @@ protected override void Arrange()
             public void Should_Valid_Instance_Id()
             {
                 A.CallTo(() => _instanceIdContextProvider.SetInstanceId(A<string>._)).MustHaveHappened();
```
blmeyers, line 148: The code you added is just calling your fake, you can just change the previous line to verify it specifically called SetInstanceId with the _instanceId value.
```suggestion
                A.CallTo(() => _instanceIdContextProvider.SetInstanceId(_instanceId)).MustHaveHappened();
```

Application/EdFi.Ods.Tests/EdFi.Ods.Api/Filters/InstanceIdContextFilterTests.cs
```diff
@@ -161,9 +163,10 @@ protected override void Arrange()
...
```
blmeyers, line 165: The code you added is just calling your fake, you can just change the previous line to verify it specifically called SetInstanceId with the _instanceId value.
```suggestion
                A.CallTo(() => _instanceIdContextProvider.SetInstanceId(_instanceId)).MustHaveHappened();
```

blmeyers (changes requested):
.Net Framework code has been completely removed, so all code conditionals (#if NETCOREAPP etc.) in new files should also go away.  In addition, some of your commits aren't signed so its still failing the cla check, you need to either squash to a single commit or interactive rebase so you can recommit each commit as signed.

Application/EdFi.Ods.Api/Middleware/EdFiOAuthAuthenticationHandler.cs
```diff
@@ -44,6 +49,29 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
...
+                        Match match = Regex.Match(instanceId, Pattern);
+                        if (match.Success)
+                        {
+                            // If we're still here, set the context value
+                            _instanceIdContextProvider.SetInstanceId(instanceId);
+                        }
+                    }
+                }
+                /* ------- */
+
```
frenchyjef, line 74: @blmeyers See notes from Mansoor. We need to get InstanceId from the route so we can use the correct Admin/Security database but since InstanceIdContextFilter is called later in the request pipeline, we had to add this code here as well. Hopefully, that is an acceptable solution.

blmeyers (approved):

PR #164, opened 2020-09-16
[ODS-4547] Fix broken integration test in the net core assemblies

jamessmckay:


Application/EdFi.Ods.WebService.Tests/Controllers/OpenApiMetadataControllerTests.cs
```diff
@@ -27,8 +27,11 @@ public async Task MetadataEndpointGetShouldBeValid()
             var json = await response.Content.ReadAsStringAsync();
 
             json.ShouldNotBeNullOrWhiteSpace();
-            Approvals.Verify(json);
```
tbmeador-dlp, line 30: This seems like it would be a stable approval test, any reason to move this back to just checking for a non-empty array?

Application/EdFi.Ods.WebService.Tests/Controllers/VersionControllerTests.cs
```diff
@@ -27,7 +28,16 @@ public async Task VersionEndpointGetShouldBeValid()
             var json = await response.Content.ReadAsStringAsync();
 
             json.ShouldNotBeNullOrWhiteSpace();
-            Approvals.Verify(json);
```
tbmeador-dlp, line 30: We'd have to update the version here when we update the constant but that's a good indication it's still working. Everything else seems like it would be pretty stable. should we keep the approval test?

tbmeador-dlp (changes requested):

Application/EdFi.Ods.WebService.Tests/Controllers/OpenApiMetadataControllerTests.cs
jamessmckay: i was seeing issues between tests that the results would be different. so i went to deserializing it to the model and checking

Application/EdFi.Ods.WebService.Tests/Controllers/VersionControllerTests.cs
jamessmckay: same reason as above, i changed it to deserializing the object and checking for the keys

tbmeador-dlp (approved):

PR #195, opened 2020-10-07
[ODS-4410] Convert all usages of \r\n in source code to use Environment.NewLine constant

tparandiev:


Application/EdFi.Ods.Common/Serialization/SerializationHelper.cs
```diff
@@ -15,7 +15,7 @@ public static class SerializationHelper
     {
         public static T DeserializeFromCompressedBase64<T>(string base64EncodedString)
         {
-            if (base64EncodedString.Contains("\n"))
+            if (base64EncodedString.Contains(Environment.NewLine))
```
blmeyers, line 18: This seems to have been incorrectly implemented in the first place.  If you Base64Encode a string and set formatting options to include line breaks, you can end up with "\r\n" in the output, but never "\n".  Even with .net standard on linux, for compatibility reasons it still outputs "\r\n".  See https://github.com/dotnet/runtime/issues/24704.  That aside, this method isn't used and seems to have some issues, so may just want to delete this method.

Application/EdFi.Ods.Common/Validation/ValidationResultExtensions.cs
```diff
@@ -43,7 +44,7 @@ public static string GetAllMessages(this ValidationResult validationResult, int
 
         private static void GetAllMessages(StringBuilder sb, ValidationResult validationResult, string indent)
         {
-            sb.AppendFormat("{0}{1}\n", indent, validationResult.ErrorMessage);
+            sb.AppendFormat($"{{0}}{{1}}{Environment.NewLine}", indent, validationResult.ErrorMessage);
```
blmeyers, line 47: Also want to document this once and let you take a pass to correct it throughout.  Anytime we're returning out to the end user we should stick with a consistent \n, rather than using Environment.NewLine. This ensures the end user parsing the response doesn't get different results based on different servers.  Environment.NewLine should be for server side uses like logging or nHibernate.

blmeyers (changes requested):
I left some higher level comments on a few examples of some areas I'd like you adjust your approach on.  Feel free to reach out to me if you have any questions, but I'll let you address those across the PR and then I'll make another pass.

blmeyers:
Also please check on these files, they seem to have been missed in this conversion:
EdFi.Ods.WebService.Tests.ApprovalTests.AggregateDependencyControllerTests
EdFi.Ods.WebService.Tests.OpenApiMetadataControllerTests
The replacing of the new lines to standardize the content is fine, but these should both be using Environment.NewLine to avoid issues running the tests on non Windows environments.

PR #216, opened 2020-10-21
[ODS-3931] Add versioning check for plugin extensions

tnunnerydl:


Application/EdFi.Ods.Common/Models/Validation/ApiModelExistsValidator.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using FluentValidation;
+
+namespace EdFi.Ods.Common.Models.Validation
+{
+    public class ApiModelExistsValidator : AbstractValidator<string>
+    {
+
```
jamessmckay, line 15:
```suggestion

```

Application/EdFi.Ods.Common/Models/Validation/ApiModelExistsValidator.cs
```diff
@@ -0,0 +1,34 @@
...
+                .Must(
+                    e =>
+                    {
+                        var metaDataFolder = Path.Combine(e, "Artifacts", "Metadata");
+                        var files = Directory
+                            .GetFiles(metaDataFolder)
+                            .Where(s => regex.IsMatch(s))
+                            .ToList();
+
+                        return files.Count > 0;
```
jamessmckay, line 30:
```suggestion
                        return Directory
                            .GetFiles(metaDataFolder)
                            .Any(s => regex.IsMatch(s));
```

jamessmckay (changes requested):

Application/EdFi.Ods.Common/Models/Validation/IsApiVersionValidValidator.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            var regex = new Regex(@"ApiModel.*.json");
+
+            _apiVersion = apiVersion;
+
+            RuleFor(extensionFolder => extensionFolder)
+                .Must(
+                    e =>
+                    {
+                        var metaDataFolder = Path.Combine(e, "Artifacts", "Metadata");
```
jamessmckay, line 29: we probably need some logging here. I think the intent is that we have an extension and that the model is valid. should be testing both of these rules as  compound?

jamessmckay (changes requested):

Application/EdFi.Ods.Common/Models/Validation/IsApiVersionValidValidator.cs
tnunnerydl: We are testing both conditions and logging is done in AssemblyLoaderHelper.

tnunnerydl: Actually, I'm not checking both conditions, only validating version here.  There's another validator that checks for extensions.

tnunnerydl: I'm not sure whey we need to make a compound validator here.  Both IsApiVersionValidValidator and IsExtensionPluginValidator are called from AssemblyLoaderHelper.  Combining these two validators into two rules in one validator will only check for valid extension assemblies.  Making the validators more granular allows us to check if the version is valid for extension, profiles and composites, right?

jamessmckay (approved):

PR #218, opened 2020-10-22
[ODS-4578] Bulk Load Client Errors

jamessmckay:


Application/EdFi.Ods.Api/Startup/OdsStartupBase.cs
```diff
@@ -270,7 +271,7 @@ void SetStaticResolvers()
 
         private void LoadPlugins()
         {
-            var pluginFolder = Plugin.Folder;
+            var pluginFolder = Path.GetFullPath(Plugin.Folder);
```
tnunnerydl, line 274: GetFullPath method is already called in PluginFinder.FindAssemliesWithPlugins.  Please remove.

Application/EdFi.Ods.Api/Container/Modules/ApplicationModule.cs
```diff
@@ -71,141 +73,177 @@ protected override void Load(ContainerBuilder builder)
...
             // Allows selected context to flow to worker Tasks (see IHttpContextStorageTransferKeys and IHttpContextStorageTransfer)
             builder.RegisterType<CallContextStorage>()
-                .As<IContextStorage>();
+                .As<IContextStorage>()
+                .SingleInstance();
 
             // Features to transfer context from HttpContext to the secondary storage in ASP.NET applications
             builder.RegisterType<HttpContextStorageTransfer>()
-                .As<IHttpContextStorageTransfer>();
+                .As<IHttpContextStorageTransfer>()
```
blmeyers, line 97: This came up from the field PR, but we've got a double register for IContextStorage happening here, which used to mean the first one would win but now means the last one will win. We're running all our requests using CallContextStorage instead of HttpContextStorage at the moment. Can we fix this with this PR?

Utilities/DataLoading/EdFi.LoadTools/Engine/ResourcePipeline/SubmitResource.cs
```diff
@@ -37,9 +37,7 @@ public async Task<ApiLoaderWorkItem> ProcessAsync(ApiLoaderWorkItem resourceWork
                     resourceWorkItem.Json, resourceWorkItem.ElementName,
                     resourceWorkItem.ResourceSchemaName, refreshToken).ConfigureAwait(false))
                 {
-                    if (response.StatusCode.Equals(HttpStatusCode.Unauthorized)
-                        && response.ReasonPhrase.Equals("Invalid token")
```
blmeyers, line 41: I'm fine with not locking down this client code for checking the ReasonPhrase, but this indicates we made a breaking change.  The OAuth spec states the ReasonPhrase here should be "Unauthorized", which is what we're doing now, but previously we were returning the OAuth error code as the ReasonPhrase.  @vimayya Is this difference important enough to correct for backwards compatibility?  If so we'll need a new ticket.

v5.0
Status: 401 Invalid token

v5.1
Status: 401 Unauthorized

blmeyers (approved):
I'd like us to revisit some of these lifecycles, but at least now nothing is transient.  I think some of those scoped could lock down further to singleton, but this is fine for now.

Utilities/DataLoading/EdFi.LoadTools/Engine/ResourcePipeline/SubmitResource.cs
vimayya: I am fine with change of reason phrase to match the spec. There are some other error description changes as well in 5.1.0 coming from .net core and overall guidance is to go by the error code rather than the description.

Application/EdFi.Ods.Api/Container/Modules/ApplicationModule.cs
jamessmckay: yes good point, I will remove the httpcontext since its not used.

blmeyers:
Issues with Context usage will be addressed in a future ticket.  Merging in so we can get templates building again.

PR #220, opened 2020-10-23
[ODS-3873] Use EdFi.Common package in ODS solution 

semalaiappan:


blmeyers (changes requested):
As per ticket, missing updates to CodeGen Utility.  Need to update the CodeGenUtility to remove the EdFi.Common project from the solution, and continue to be dependent on EdFi.Ods.Common for now (with its package dependency on EdFi.Suite3.Common).  Then make sure the built CodeGen is pulled in on your Implementation branch.

blmeyers (approved):

PR #221, opened 2020-10-26
[ODS-4633] Vendor key is shared between different requests

gmcelhanon:
Restores the `CallContext`'s use of `AsyncLocal` without the value changed handler that was previously used to "copy" the context -- this ended up corrupting the default context used for new `ExecutionContext` instances, causing the "cross-contamination" of context on other requests.

The change also fixes the original problem with `ApiKeyContext` being lost. The context was actually lost _immediately_ upon return from the "handler" method in which it is set since context only flows _down_, not _up_ the async call stack. By setting it in the WebAPI authentication middleware (which unfortunately does not have a seam for that particular behavior and thus requires a wholesale replacement with necessary enhancement), the context is now available down the call chain for the duration of the request.

This also relieves us from having to try to use Autofac-specific behaviors to try to isolate the context for the current request, allowing us to revert to freely using singletons where it makes sense (where the components do not maintain any state).


Application/EdFi.Ods.Api/Middleware/EdFiApiAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (defaultAuthenticate != null)
+            {
+                var result = await context.AuthenticateAsync(defaultAuthenticate.Name);
+
+                if (result?.Principal != null)
+                {
+                    context.User = result.Principal;
+
+                    // NOTE: For our use case we set the api key context into the call context storage. The rest of this code
+                    // is the default implementation and may need to be update on version updates.
```
blmeyers, line 71: Link to the source for future updates: https://github.com/dotnet/aspnetcore/blob/v3.1.9/src/Security/Authentication/Core/src/AuthenticationMiddleware.cs is what I'm assuming this class was sourced from?  We seem to be not quite identical as well.

blmeyers (changes requested):

Application/EdFi.Ods.Api/Middleware/EdFiApiAuthenticationMiddleware.cs
jamessmckay: yes geoff decompiled the assembly

blmeyers: We should match source then, we're both Apache so we can share the source code.

blmeyers (approved):

PR #222, opened 2020-10-28
[ODS-4592] Duplicates from Security Views

syednoorulla:
… the selection.


gmcelhanon (changes requested):
As a general note when authoring database change scripts -- you should never use the initial creation script as the source for a new change script. Instead, you must start with the **current** state of the authorization views from the most recent ODS database. Your original PR, which included `ALTER` blocks for views unaffected by this ticket was actually reintroducing a bug because the original change script was not correct.

In any case, there are a few changes I've requested.

gmcelhanon:
Also @syednoorulla, please rename the file to `1150-UpdatedAuthViewsToPreventDuplicatesFromStaffEdOrgAssignmentsAndEmployments.sql`.

gmcelhanon:
Closing PR to produce a clean history for the changes.

PR #247, opened 2020-12-07
[ODS-4664] Dynamic Plugin Validation based on assemblyMetadata.json file in FindPluginAssemblies method

semalaiappan:


Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -142,35 +147,34 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
-                    var validationResult = validator.ValidateObject(extensionFolder);
-
-                    if (!validationResult.Any())
+                    string[] assemblyMetadataPath = Directory.GetFiles(
+                        extensionFolder, AssemblyMetadataSearchString, SearchOption.AllDirectories);
+                        
+                    if (assemblyMetadataPath.Length>0 && IsExtensionInAssemblyModelType(assemblyMetadataPath[0]))
                     {
-                        yield return assembly.Location;
+                        var validator = GetValidator();
```
gmcelhanon, line 159: For clarity, please rename this method to `GetExtensionValidator` since the validations only apply to extension assemblies and _definitely_ not the `else` condition for the profile assemblies.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
...
+            var assemblyMetadata = Load(assemblyMetadataPath);
+
+            return assemblyMetadata.AssemblyModelType.EqualsIgnoreCase(TemplateSetConventions.Profile);
+        }
+
+        public static AssemblyMetadata Load(string fileFullName)
+       {
+           if (!File.Exists(fileFullName))
+           {
+               throw new FileNotFoundException($"Non-existent file path provided. Expected location {fileFullName}.");
```
gmcelhanon, line 227: Wrap all message parameters in single quotes (i.e. add quotes around `{fileFullName}').

Suggest changing the message to `$"Assembly metadata file '{fileFullName}' not found."`

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
...
+           }
+
+           _logger.Debug($"Deserializing object type {typeof(AssemblyMetadata)} from file {fileFullName}.");
```
gmcelhanon, line 230: Add single quotes around parameterized values in messages.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
...
+
+           using (StreamReader file = File.OpenText(fileFullName))
```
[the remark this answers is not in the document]
gmcelhanon: You might consider renaming the parameter to this method from `fileFullName` to `assemblyMetadataPath` and rewriting this logic more simply as:
```csharp
string json = File.ReadAllText(assemblyMetadataPath);
return JsonConvert.DeserializeObject<AssemblyMetadata>(json);
```

gmcelhanon (changes requested):
This represents a first round of code changes. Once these changes are made, then I'll review and test it for desired functionality.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -13,18 +13,25 @@
...
+using Newtonsoft.Json;
 
 namespace EdFi.Ods.Api.Helpers
 {
     public static class AssemblyLoaderHelper
     {
         private static readonly ILog _logger = LogManager.GetLogger(typeof(AssemblyLoaderHelper));
-
+        private const string AssemblyMetadataSearchString = "assemblyMetadata.json";
         public static void LoadAssembliesFromExecutingFolder(bool includeFramework = false)
```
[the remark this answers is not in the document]
semalaiappan: yes .  I updated it . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -142,35 +147,34 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
+                if (Directory.Exists(extensionFolder))
                 {
-                    string extensionFolder = Path.GetDirectoryName(assemblyPath);
-                    var validationResult = validator.ValidateObject(extensionFolder);
-
-                    if (!validationResult.Any())
+                    string[] assemblyMetadataPath = Directory.GetFiles(
+                        extensionFolder, AssemblyMetadataSearchString, SearchOption.AllDirectories);
+                        
+                    if (assemblyMetadataPath.Length>0 && IsExtensionInAssemblyModelType(assemblyMetadataPath[0]))
```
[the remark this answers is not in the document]
semalaiappan: I removed this logic . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -142,35 +147,34 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
                     {
-                        yield return assembly.Location;
+                        var validator = GetValidator();
+                        var validationResult = validator.ValidateObject(extensionFolder);
+                        if (!validationResult.Any())
```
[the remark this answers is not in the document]
semalaiappan: yes I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -183,6 +187,17 @@ static FluentValidationObjectValidator GetValidator()
                     });
             }
 
+            bool IsPluginFolderNameSupplied()
+            {
+                if (!string.IsNullOrEmpty(pluginFolder) && !string.IsNullOrWhiteSpace(pluginFolder))
```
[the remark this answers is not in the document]
semalaiappan: Agreed . removed it . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
...
+           _logger.Debug($"Deserializing object type {typeof(AssemblyMetadata)} from file {fileFullName}.");
+
+           using (StreamReader file = File.OpenText(fileFullName))
+           {
+               JsonSerializer serializer = new JsonSerializer();
+               return (AssemblyMetadata)serializer.Deserialize(file, typeof(AssemblyMetadata));
+           }
+       }
+
+
```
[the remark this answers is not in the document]
semalaiappan: I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
...
+           {
+               throw new FileNotFoundException($"Non-existent file path provided. Expected location {fileFullName}.");
+           }
+
+           _logger.Debug($"Deserializing object type {typeof(AssemblyMetadata)} from file {fileFullName}.");
+
+           using (StreamReader file = File.OpenText(fileFullName))
+           {
+               JsonSerializer serializer = new JsonSerializer();
+               return (AssemblyMetadata)serializer.Deserialize(file, typeof(AssemblyMetadata));
```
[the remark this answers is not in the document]
semalaiappan: Removed this logic . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
             }
         }
 
+        private static bool IsExtensionInAssemblyModelType(string assemblyMetadataPath)
+        {
+            var assemblyMetadata = Load(assemblyMetadataPath);
```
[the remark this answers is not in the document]
semalaiappan: yes ,  I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
semalaiappan: yes ,  I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
semalaiappan: yes ,  I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
semalaiappan: yes ,  I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -142,35 +147,34 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
                 {
                     continue;
                 }
-
-                if (assemblyPath.Contains("Extensions"))
+                
+                string extensionFolder = Path.GetDirectoryName(assemblyPath);
+                if (Directory.Exists(extensionFolder))
```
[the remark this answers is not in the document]
semalaiappan: yes ,  I removed. Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
semalaiappan: yes ,  I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,6 +206,37 @@ static bool HasPlugin(Assembly assembly)
...
+        }
+
+        private static bool IsProfileInAssemblyModelType(string assemblyMetadataPath)
+        {
+            var assemblyMetadata = Load(assemblyMetadataPath);
+
+            return assemblyMetadata.AssemblyModelType.EqualsIgnoreCase(TemplateSetConventions.Profile);
+        }
+
+        public static AssemblyMetadata Load(string fileFullName)
```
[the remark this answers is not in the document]
semalaiappan: yes ,  I updated . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
semalaiappan: I implemented this logic and test it also . Thanks

semalaiappan:
@gmcelhanon  - I fixed all review comments .  Ready for review  again

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -142,37 +148,41 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
                 {
                     continue;
                 }
+                
+                string extensionFolder = Path.GetDirectoryName(assemblyPath);
+                string assemblyMetadataPath = Directory.GetFiles(
```
gmcelhanon, line 153: The requirement on [the ticket](https://tracker.ed-fi.org/browse/ODS-4664) is for the _assemblyMetadata.json_ file to be an embedded resource in the plugin assembly.

You'll need to use the `Assembly` class' `GetManifestResourceNames` and `GetManifestResourceStream` methods to get the resource from the assembly. Then you can use a `StreamReader` and its `ReadToEnd` method (see documentation [here](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readtoend?view=net-5.0)) to get the JSON for deserialization.

This will also require changes to the projects in the _Ed-Fi-Extensions repository_, and also the sample Profiles project (?).

Application/EdFi.Ods.Api/Models/AssemblyData.cs
```diff
@@ -0,0 +1,25 @@
+﻿// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System;
+
+namespace EdFi.Ods.Api.Models
+{
+    public class AssemblyData
```
gmcelhanon, line 10: This class does not appear to be used and looks maybe like it was just copied in from the CodeGen utility.

Please remove it.

Application/EdFi.Ods.Api/Models/AssemblyMetadata.cs
```diff
@@ -0,0 +1,12 @@
+﻿// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.Ods.Api.Models
+{
+    public class AssemblyMetadata
```
gmcelhanon, line 8: The location of this implies that it is part of the API's models, when in fact it is just be used as a object for deserializing some custom assembly metadata for internal purposes during initialization. It really has no general purpose.

Please remove this file, but add the class as a `private class` at the bottom of _AssemblyLoaderHelper.cs_ file, like so:
```csharp
        ...

        private class PluginAssemblyLoadContext : AssemblyLoadContext
        {
            public PluginAssemblyLoadContext()
                : base(isCollectible: true) { }
        }
        
        private class AssemblyMetadata
        {
            public string AssemblyModelType { get; set; }
        }
    }
}
```

gmcelhanon (changes requested):
I missed noticing the requirement for the assembly metadata to be embedded into the assembly on my first pass review, so there are some additional changes that need to be made for that.

Application/EdFi.Ods.Api/Conventions/TemplateSetConventions.cs
```diff
@@ -0,0 +1,13 @@
+﻿// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+namespace EdFi.Ods.Api.Conventions
+{
+    public static class TemplateSetConventions
```
gmcelhanon, line 8: "TemplateSet" is terminology copied in from the CodeGen utility. I think `PluginConventions` would be a better name for this class in this context.

Application/EdFi.Ods.Api/Models/AssemblyData.cs
semalaiappan: Fixed it . Thanks

Application/EdFi.Ods.Api/Conventions/TemplateSetConventions.cs
semalaiappan: Fixed it . Thanks

semalaiappan:
@gmcelhanon . Ready for review again . Thanks

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,7 +203,38 @@ static bool HasPlugin(Assembly assembly)
...
+                throw new FileNotFoundException($"Assembly metadata file '{assembly.GetManifestResourceNames()}' not found for this project.");
+            }
+            var stream = assembly.GetManifestResourceStream(resourceName);
+
+            using (StreamReader reader = new StreamReader(stream))
+            {
+                _jsonFile = reader.ReadToEnd(); 
+            }
+
+            _logger.Debug($"Deserializing object type {typeof(AssemblyMetadata)} from file '{resourceName}'.");
```
gmcelhanon, line 232: Missing single quotes around type parameter in message.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,7 +203,38 @@ static bool HasPlugin(Assembly assembly)
...
+        }
+
+        private static AssemblyMetadata LoadAssemblyMetadata(Assembly assembly)
+        {
+            var resourceName = assembly.GetManifestResourceNames()
+                .SingleOrDefault(p => p.EndsWith(AssemblyMetadataSearchString));
+            
+            if (resourceName == null)
+            {
+                throw new FileNotFoundException($"Assembly metadata file '{assembly.GetManifestResourceNames()}' not found for this project.");
```
gmcelhanon, line 223: Usage of a `FileNotFoundException` is no longer appropriate here since we're looking for an embedded resource, not a file. Use `Exception` instead.

The parameter in the error message is now using `assembly.GetManifestResourceNames()` which is wrong on 2 accounts:

1. That method returns a `string[]`, which when embedded in the message like this will result in the literal text `System.String[]`, not the members of the array in a readable format.
2. The names of all the resources that are actually in the assembly isn't right for this message. We should be informing on which one is _missing_. This would be the value stored in the `AssemblyMetadataSearchString` constant.

Fix the message, and change the use of the term "project" to be "assembly", as that is what is actually be searched.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,7 +203,38 @@ static bool HasPlugin(Assembly assembly)
...
+            }
+            var stream = assembly.GetManifestResourceStream(resourceName);
```
gmcelhanon, line 225: You need to add blank lines before and after blocks of code. Insert a blank line before `var stream` here.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -127,52 +135,45 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
+            
             var assemblies = Directory.GetFiles(pluginFolder, "*.dll", SearchOption.AllDirectories);
 
             var pluginFinderAssemblyContext = new PluginAssemblyLoadingContext();
 
             foreach (var assemblyPath in assemblies)
             {
                 var assembly = pluginFinderAssemblyContext.LoadFromAssemblyPath(assemblyPath);
-
+
```
gmcelhanon, line 146: Remove spurious whitespace changes.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -13,18 +13,26 @@
...
+using Newtonsoft.Json;
 
 namespace EdFi.Ods.Api.Helpers
 {
     public static class AssemblyLoaderHelper
     {
         private static readonly ILog _logger = LogManager.GetLogger(typeof(AssemblyLoaderHelper));
-
+        private const string AssemblyMetadataSearchString = "assemblyMetadata.json";
+        private static string _jsonFile ;
```
gmcelhanon, line 35: You've got a space before the final `;`, but more importantly, why is this a static variable?

This is potentially dangerous as it means there is 1 copy of this variable for all threads in the application to share. In the event multiple calls were made in parallel to this class (which in this case I don't think will actually happen), the threads would be overwriting each others values. It's just not a good practice.

Also, this value is only used inside of one method. Please make it a local variable inside of that method.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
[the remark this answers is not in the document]
gmcelhanon: After these latest changes, you lost the blank line. Please make sure that after moving the `_jsonFile` variable to the method, that you leave a blank line in here before the method declaration.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -127,52 +135,45 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
-
-            pluginFinderAssemblyContext.Unload();
-
-            bool IsPluginFolderNameSupplied()
-            {
-                if (!string.IsNullOrEmpty(pluginFolder) && !string.IsNullOrWhiteSpace(pluginFolder))
+                else if (IsProfileAssembly(assemblyMetadata))
                 {
-                    return true;
+                     yield return assembly.Location;
```
gmcelhanon, line 171: This is indented 5 spaces. Please only indent using 4 spaces.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -127,52 +135,45 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
-                        _logger.Warn($"Assembly: {assembly.GetName()} - {string.Join(",", validationResult)}");
-                    }
+                if (IsExtensionAssembly(assemblyMetadata))
+                {
+                     var validator = GetExtensionValidator();
+                     var validationResult = validator.ValidateObject(extensionFolder);
+
+                     if (!validationResult.Any())
+                     {
+                          yield return assembly.Location;
```
gmcelhanon, line 162: This is indented 5 spaces. Please only indent using 4 spaces.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -127,52 +135,45 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
-                    {
-                        yield return assembly.Location;
-                    }
-                    else
-                    {
-                        _logger.Warn($"Assembly: {assembly.GetName()} - {string.Join(",", validationResult)}");
-                    }
+                if (IsExtensionAssembly(assemblyMetadata))
+                {
+                     var validator = GetExtensionValidator();
```
gmcelhanon, line 157: This whole block is indented 5 spaces. Please only indent using 4 spaces.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -127,52 +135,45 @@ public static IEnumerable<string> FindPluginAssemblies(string pluginFolder, bool
...
+                     var validationResult = validator.ValidateObject(extensionFolder);
+
+                     if (!validationResult.Any())
+                     {
+                          yield return assembly.Location;
+                     }
+                     else
+                     {
+                          _logger.Warn($"Assembly: {assembly.GetName()} - {string.Join(",", validationResult)}");
```
gmcelhanon, line 166: Indented 5 spaces.

Application/EdFi.Ods.Api/Models/AssemblyMetadata.cs
gmcelhanon: @semalaiappan Looks like you missed this one. Please move the `AssemblyMetadata` class as described in this comment, and make it private.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,7 +203,38 @@ static bool HasPlugin(Assembly assembly)
...
+            var resourceName = assembly.GetManifestResourceNames()
+                .SingleOrDefault(p => p.EndsWith(AssemblyMetadataSearchString));
+            
+            if (resourceName == null)
+            {
+                throw new FileNotFoundException($"Assembly metadata file '{assembly.GetManifestResourceNames()}' not found for this project.");
+            }
+            var stream = assembly.GetManifestResourceStream(resourceName);
+
+            using (StreamReader reader = new StreamReader(stream))
```
gmcelhanon, line 227: Use `var` here.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,7 +203,38 @@ static bool HasPlugin(Assembly assembly)
...
+            {
+                _jsonFile = reader.ReadToEnd(); 
+            }
+
+            _logger.Debug($"Deserializing object type {typeof(AssemblyMetadata)} from file '{resourceName}'.");
+
+            return JsonConvert.DeserializeObject<AssemblyMetadata>(_jsonFile);
+        }
+
+       private class PluginAssemblyLoadingContext : AssemblyLoadContext
```
gmcelhanon, line 237: This is only indented 3 spaces instead of 4.

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
```diff
@@ -191,7 +203,38 @@ static bool HasPlugin(Assembly assembly)
...
+
+        private static bool IsProfileAssembly(AssemblyMetadata assemblyMetadata)
+        {
+            return assemblyMetadata.AssemblyModelType.EqualsIgnoreCase(PluginConventions.Profile);
+        }
+
+        private static AssemblyMetadata LoadAssemblyMetadata(Assembly assembly)
+        {
+            var resourceName = assembly.GetManifestResourceNames()
+                .SingleOrDefault(p => p.EndsWith(AssemblyMetadataSearchString));
```
gmcelhanon, line 219: Now with the way this is written (to search for any embedded resources that end with "assemblyMetadata.json"), there is a chance (if the assembly author put multiple copies of the file at different locations) that we could end up with more than 1 matching embedded resource. In this case `SingleOfDefault` will throw an exception.

Let's change this to `FirstOrDefault`, so that we use the first one found (they _should_ only include 1, but you never know).

gmcelhanon (changes requested):

Application/EdFi.Ods.Api/Helpers/AssemblyLoaderHelper.cs
gmcelhanon: You should also probably change the message to use "embedded resource" instead of "file" now.

gmcelhanon (approved):

PR #249, opened 2020-12-08
[ODS-4022] availableChangeQueries OpenApi Spec definition is incorrect

raghuramakrishna:


gmcelhanon (changes requested):
Neither the API nor the Open API metadata should have the "Id" suffix.

The Open API metadata should match the casing of the actual JSON response, and should not have the "Id" suffix. There should be no changes to the API itself.

raghuramakrishna:
@gmcelhanon I have changed the code as mentioned. Please review.

gmcelhanon (approved):

PR #267, opened 2021-02-05
[ODS-4714] PUT description incorrect in SwaggerUI/OpenAPI Specification

raghuramakrishna:
…ther it supports update


Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
```diff
@@ -265,6 +263,21 @@ private Operation CreatePutByIdOperation(OpenApiMetadataPathsResource openApiMet
                 parameters = CreatePutParameters(openApiMetadataResource),
                 responses = OpenApiMetadataDocumentHelper.GetWriteOperationResponses(HttpMethod.Put)
             };
+
+            if(openApiMetadataResource.Resource.Entity.Identifier.IsUpdatable)
```
gmcelhanon, line 267: Per our coding standards, you need a space between the `if` and the opening `(`.

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
```diff
@@ -349,7 +362,7 @@ private Operation CreatePostOperation(OpenApiMetadataPathsResource openApiMetada
...
                 description =
-                    "The POST operation can be used to create or update resources. In database terms, this is often referred to as an \"upsert\" operation (insert + update). Clients should NOT include the resource \"id\" in the JSON body because it will result in an error (you must use a PUT operation to update a resource by \"id\"). The web service will identify whether the resource already exists based on the natural key values provided, and update or create the resource appropriately.",
+                    "The POST operation can be used to create or update resources. In database terms, this is often referred to as an \"upsert\" operation (insert + update). Clients should NOT include the resource \"id\" in the JSON body because it will result in an error. The web service will identify whether the resource already exists based on the natural key values provided, and update or create the resource appropriately. It is recommended to use POST for both create and update except while updating natural key of a resource in which case PUT operation could be used.",
```
gmcelhanon, line 365: Change the verbiage here slightly, replacing the ending of `could be used` with `must be used`.

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
gmcelhanon: Rather than inline this logic after the object initialization, please follow the example of how some of the other properties are initialized here. Refactor this logic into a separate method (i.e. `GetDescription`, passing the `openApiMetadataResource` as the argument) and call that in the initializer, as is done for the `parameters` property (for example).

Application/EdFi.Ods.Standard/Artifacts/Metadata/DomainMetadata.xml
```diff
@@ -1,1106 +0,0 @@
-<?xml version="1.0" encoding="utf-8" ?>
```
gmcelhanon, line 1: Did you do anything to check to make absolutely sure that this isn't used anywhere and can be removed now?

gmcelhanon (changes requested):

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
```diff
@@ -246,16 +246,14 @@ private IList<Parameter> CreateQueryParameters(bool isCompositeContext)
 
         private Operation CreatePutByIdOperation(OpenApiMetadataPathsResource openApiMetadataResource)
         {
-            return new Operation
+            var op = new Operation
```
gmcelhanon, line 249: After refactoring the logic (as described below), revert this line and return the operation directly.

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
raghuramakrishna: updated

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
raghuramakrishna: updated

Application/EdFi.Ods.Standard/Artifacts/Metadata/DomainMetadata.xml
raghuramakrishna: Updated techdocs. Added XML file back in project.

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
raghuramakrishna: updated

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
```diff
@@ -253,20 +253,37 @@ private Operation CreatePutByIdOperation(OpenApiMetadataPathsResource openApiMet
...
                 operationId = $"put{openApiMetadataResource.Name}",
                 deprecated = openApiMetadataResource.IsDeprecated,
                 consumes = new[]
                 {
                     _contentTypeStrategy.GetOperationContentType(openApiMetadataResource, ContentTypeUsage.Writable)
                 },
                 parameters = CreatePutParameters(openApiMetadataResource),
-                responses = OpenApiMetadataDocumentHelper.GetWriteOperationResponses(HttpMethod.Put)
+                responses = OpenApiMetadataDocumentHelper.GetWriteOperationResponses(HttpMethod.Put),
+                isUpdatable = CheckIsUpdatable(openApiMetadataResource)
```
gmcelhanon, line 266: I found the name on this method misleading here.

On first read, it looked like you were simply checking the `IsUpdatable` property - effectively making it a function that should have been reused in the `GetDescription` method below as essentially a shorthand for the code of `openApiMetadataResource.Resource.Entity.Identifier.IsUpdatable`.

However, as I started writing comments I realized that the `CheckIsUpdatable` isn't so much about checking whether it's updatable, but it's actually preparing the flag for assignment to the Open API model as either `true` or `null` because it represents custom metadata and the JSON serializer will drop the property from the serialized output if it's `null`. That is not what the method name implies at all with the chosen verb "Check". Choosing good names is very important for making code maintainable.

Since you are actually preparing a specific model value for custom metadata in this method, please rename it to `GetIsUpdatableCustomMetadataValue`.

Then, shorten the implementation to this:
```
private bool? GetIsUpdatableCustomMetadataValue(OpenApiMetadataPathsResource openApiMetadataResource)
{
    return openApiMetadataResource.Resource.Entity.Identifier.IsUpdatable ? (bool?) true : null;
}
```

gmcelhanon (changes requested):

Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataPathsFactory.cs
jamessmckay: done

gmcelhanon (approved):

PR #271, opened 2021-02-11
[ODS-4792] Code Gen Build Isn't Running Tests

tbmeador-dlp:


Utilities/CodeGeneration/EdFi.Ods.CodeGen/Providers/Impl/LegacyExtensionAssemblyDataProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+                    SearchOption.AllDirectories)
+                .Select(
+                    x =>
+                    {
+                        string assemblyName = _assemblyDataHelper.GetAssemblyName(x);
+
+                        string schemaName = _domainModelsDefinitionsProvidersByProjectName[assemblyName]
+                            .GetDomainModelDefinitions()
+                            .SchemaDefinition
+                            .LogicalName;
```
gmcelhanon, line 60: Same question with the use of the LogicalName. Shouldn't this usage (and really anywhere else I might have missed in this review) be converted appropriately to `ProperCaseName` using the extension method mentioned previously?

gmcelhanon (changes requested):

jamessmckay (approved):

PR #282, opened 2021-02-26
[ODS-4843] Pre release tasks for ODS/API 5.2.0

raghuramakrishna:


blmeyers (changes requested):
Changes are fine, but we need to update the backup packages or initdev will fail since they're out of sync now.

tbmeador-dlp (approved):

PR #287, opened 2021-03-04
[ODS-4852] Minimal Template Builds fails sometimes after merge of ODS-4788

tbmeador-dlp:
DefaultRequestHeaders is not thread safe
see: https://github.com/dotnet/runtime/issues/1500

### DIFF
--- a/Utilities/DataLoading/EdFi.LoadTools/ApiClient/TokenRetriever.cs
+++ b/Utilities/DataLoading/EdFi.LoadTools/ApiClient/TokenRetriever.cs
@@ -38,17 +38,21 @@ namespace EdFi.LoadTools.ApiClient
 
         public async Task<BearerToken> ObtainNewBearerToken()
         {
+            var requestMessage = new HttpRequestMessage
+            {
+                Method = HttpMethod.Post,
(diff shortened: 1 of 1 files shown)


Utilities/DataLoading/EdFi.LoadTools/ApiClient/TokenRetriever.cs
```diff
@@ -38,17 +38,21 @@ public TokenRetriever(IOAuthTokenConfiguration configuration)
 
         public async Task<BearerToken> ObtainNewBearerToken()
         {
+            var requestMessage = new HttpRequestMessage
+            {
+                Method = HttpMethod.Post,
+                Content = new StringContent("Grant_type=client_credentials", Encoding.UTF8, "application/x-www-form-urlencoded"),
+                RequestUri = new Uri(_configuration.Url)
+            };
+
```
jamessmckay, line 47: I am was a little concerned about this change, however after researching, I am more comfortable with the change. The reason for the change is to reset the header values and force them to change. c.f. https://stebet.net/httpclient-best-practices/

jamessmckay (approved):

PR #300, opened 2021-03-17
[AA-853] Enhance AA/ODS decoupling by using XSDs served up by ODS-4782

CSR2017:
@jamessmckay @vimayya @plioi 
Please help review and merge the changes.
Changes:
1. Introduced new configuration parameter AdminAppSchemaAlreadyDownloaded to control the Xsd download from AdminApp
2. Refactored Xsd download steps, so can be called from Run and ValidateXmlFile methods.
3. On Run method calling LoadXsdFiles when AdminAppSchemaAlreadyDownloaded  is false.

Thanks

### DIFF
--- a/Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
+++ b/Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
@@ -29,6 +29,7 @@ namespace EdFi.LoadTools.BulkLoadClient.Application
 
         public string WorkingFolder { get; set; }
 
+        public bool Metadata { get; set; }
 
         public string MetadataUrl { get; set; }
 
@@ -61,7 +62,7 @@ namespace EdFi.LoadTools.BulkLoadClient.Application
 
         bool IApiMetadataConfiguration.Force
         {
-            get => ForceMetadata;
+            get => Metadata;
         }
 
         string IApiMetadataConfiguration.Folder
@@ -120,12 +121,13 @@ namespace EdFi.LoadTools.BulkLoadClient.Application
         {
             get => DoNotValidateXml;
         }
-
         public string Extension { get; set; }
 
         public string XsdMetadataUrl { get; set; }
 
         public bool ForceMetadata { get; set; }
+
+        public bool AdminAppSchemaAlreadyDownloaded { get; set; }
 
         public static BulkLoadClientConfiguration Create(IConfiguration configuration)
         {
@@ -138,16 +140,16 @@ namespace EdFi.LoadTools.BulkLoadClient.Application
                 : Path.GetFullPath(configuration.GetValue<string>("Folders:Working"));
 
             var xsdFolder = string.IsNullOrEmpty(configuration.GetValue<string>("Folders:Xsd"))
-                ? Path.Combine(workingFolder, "Schemas")
+                ? Path.Combine(workingFolder, "xsd")
                 : Path.GetFullPath(configuration.GetValue<string>("Folders:Xsd"));
 
             return new BulkLoadClientConfiguration
             {
                 ConnectionLimit = configuration.GetValue("Concurrency:ConnectionLimit", 100),
                 DataFolder = configuration.GetValue("Folders:Data", currentDirectory),
-                DoNotValidateXml = !configuration.GetValue<bool>("ValidateSchema"),
+                DoNotValidateXml = configuration.GetValue<bool>("ValidateSchema"),
                 Profile = configuration.GetValue<string>("OdsApi:Profile"),
-                ForceMetadata = configuration.GetValue<bool>("ForceMetadata"),
+                Metadata = configuration.GetValue<bool>("ForceMetadata"),
                 Retries = configuration.GetValue("Concurrency:MaxRetries", 3),
                 IncludeStats = configuration.GetValue<bool>("IncludeStats"),
                 OAuthKey = configuration.GetValue<string>("OdsApi:Key"),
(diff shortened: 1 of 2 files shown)


Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
```diff
@@ -165,7 +167,8 @@ public static BulkLoadClientConfiguration Create(IConfiguration configuration)
                 XsdMetadataUrl = configuration.GetValue<string>("OdsApi:XsdMetadataUrl"),
                 Extension = configuration.GetValue<string>("OdsApi:Extension"),
                 ApiMode =  apiMode,
-                ForceMetadata = configuration.GetValue<bool>("ForceMetadata")
+                ForceMetadata = configuration.GetValue<bool>("ForceMetadata"),
+                AdminAppSchemaAlreadyDownloaded = configuration.GetValue<bool>("AdminAppSchemaAlreadyDownloaded")
```
vimayya, line 171: @CSR2017 Why couldn't we use ForceMetadata configuration here? I think that should guide the download instead of adding a new configuration. Similar process is done for json metadata https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/main/Utilities/DataLoading/EdFi.LoadTools/ApiClient/SwaggerMetadataRetriever.cs#L41 cc @jamessmckay

vimayya (changes requested):

Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
tbmeador-dlp: I think one of problems here is that the downloader is pretty aggressive. Even if you provide it with an existing folder of schema files to work off of, if it finds the xsd metadata url it will redownload the files even when ForceMetdata is false.

I was trying to work towards a solution where if you explicitly provide the folder and it is not empty it will attempt to use that as its schema source.

vimayya: > I think one of problems here is that the downloader is pretty aggressive. Even if you provide it with an existing folder of schema files to work off of, if it finds the xsd metadata url it will redownload the files even when ForceMetdata is false.
> 
> I was trying to work towards a solution where if you explicitly provide the folder and it is not empty it will attempt to use that as its schema source.

@tbmeador-dlp is that change coming in ODS-4858?

tbmeador-dlp: @vimayya yes, I was hoping to include it with that ticket

CSR2017: @vimayya @tbmeador-dlp 
Updated the code based on ODS-4858 changes. Need review and merge after merging https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/298/files

Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/LoadProcess.cs
```diff
@@ -173,24 +183,23 @@ private static IEnumerable<string> GetFilesToImport(BulkLoadClientConfiguration
...
                         }
-
-                        reader.Close();
                     }
 
-                    fileStream.Close();
+                    reader.Close();
                 }
+
+                fileStream.Close();
```
jamessmckay, line 202: so the recommendation is not call close, and let the dispose do the work.
```
Closes the current stream and releases any resources (such as sockets and file handles) associated with the current stream. Instead of calling this method, ensure that the stream is properly disposed.(Inherited from Stream)
```
I see value in calling `reader.FlushAsync()` instead.

Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/LoadProcess.cs
```diff
@@ -37,14 +36,7 @@ public static async Task<int> Run(IConfiguration configuration)
...
 
-                using var xsdFilesRetriever = new XsdFilesRetriever(
-                    bulkLoadClientConfiguration,
-                    new XsdMetadataInformationProvider(),
-                    new XsdMetadataFilesProvider(),
-                    new RemoteFileDownloader(),
-                    odsVersionInformation);
-
-                await xsdFilesRetriever.DownloadXsdFilesAsync();
+                await LoadXsdFiles(bulkLoadClientConfiguration, odsVersionInformation);
```
jamessmckay, line 39: should we be using ConfigureAwait(false)? Please confirm that is what we used everywhere else.

jamessmckay (changes requested):

jamessmckay (approved):

Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
```diff
@@ -138,16 +138,16 @@ public static BulkLoadClientConfiguration Create(IConfiguration configuration)
...
-                ? Path.Combine(workingFolder, "Schemas")
+                ? Path.Combine(workingFolder, "xsd")
                 : Path.GetFullPath(configuration.GetValue<string>("Folders:Xsd"));
 
             return new BulkLoadClientConfiguration
             {
                 ConnectionLimit = configuration.GetValue("Concurrency:ConnectionLimit", 100),
                 DataFolder = configuration.GetValue("Folders:Data", currentDirectory),
-                DoNotValidateXml = !configuration.GetValue<bool>("ValidateSchema"),
+                DoNotValidateXml = configuration.GetValue<bool>("ValidateSchema"),
```
tbmeador-dlp, line 148: most of the changes to this file seem to be reverting recent changes. was that intentional?

tbmeador-dlp (changes requested):

jamessmckay (approved):

vimayya (approved):

PR #305, opened 2021-03-24
[ODS-4903] Load tool - year specific mode issue

jamessmckay:


Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
```diff
@@ -171,7 +171,7 @@ string ResolvedUrl(string url)
             {
                 return apiMode == ApiMode.YearSpecific
                     ? Regex.Replace(
-                        url, @"(?<year>\d{4})\b", configuration.GetValue<string>("OdsApi:SchoolYear"), RegexOptions.None)
+                        url, @"\/(?<year>\b\d{4}\b)\/", configuration.GetValue<string>("OdsApi:SchoolYear"), RegexOptions.None)
```
blmeyers, line 174: This applies to all the regex usage.  You're requiring a trailing slash now on this url, and the api endpoint doesn't actually generate one, so this regex doesn't work to match for several of the urls.

This is also still aggressively catching 4 digits earlier in the url, but it is definitely a lot less aggressive than it was. If we can get it handling without the trailing slash that's probably fine for now, but eventually we'd probably want to handle this by only replacing the last match and not all matches.

See https://api-stage.ed-fi.org/YearSpecific_v5.2/api/.

blmeyers (changes requested):

Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
```diff
@@ -171,7 +171,9 @@ string ResolvedUrl(string url)
             {
                 return apiMode == ApiMode.YearSpecific
                     ? Regex.Replace(
-                        url, @"(?<year>\d{4})\b", configuration.GetValue<string>("OdsApi:SchoolYear"), RegexOptions.None)
+
+                        // https://regex101.com/r/lgugB1/1
```
blmeyers, line 175: Can you update this to https://regex101.com/r/KywmUK/1 everywhere to reflect your changes to the Regex replace.

blmeyers (changes requested):
Minor documentation update, but seems to be working with local testing.

Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfiguration.cs
tbmeador-dlp: done

blmeyers (approved):

PR #313, opened 2021-04-26
[ODS-4900] Cleanup Unused References in ODS/API Solution

semalaiappan:


tbmeador-dlp (changes requested):
if these aren't used anywhere else can we go ahead and removes it from the notices.md files?

https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/af335643382663ea430804d4072b330c51b7c487/NOTICES.md#L14-L44


https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/33939d3772a667515fe107db0b26bb7f06e0c289/NOTICES.md#L14-L44

tbmeador-dlp (approved):

PR #321, opened 2021-06-29
[ODS-4989] Bulk Load Client is not working 2022 Year Instance

tnunnerydl:
…d oauthurl.


tbmeador-dlp (changes requested):
Testing with the year parameter appears to be working correctly now. If I try to pass values without the year to a year specific api previous versions would inform me that the year is a required parameter. With these changes it appears to just be replacing the year in the url with nothing and then failing to load. This is the validation error I get with previous versions: https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/d03c931840cfd1939d0f4f67f8b4159c29eab893/Utilities/DataLoading/EdFi.LoadTools/BulkLoadClient/Application/BulkLoadClientConfigurationValidator.cs#L96
It seems to be skipping this validation message and going directly to downloading the xsds with the year removed from the url.

tbmeador-dlp (approved):

PR #331, opened 2021-07-28
[ODS-5007] PUT doesn't return location header in v5.1+

tbmeador-dlp:


Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
```diff
@@ -322,34 +326,35 @@ private EntityTagHeaderValue GetEtag(string etagValue)
...
 
-        protected string GetResourceUrl(Guid id)
+        protected string GetApplicationUrl()
         {
-            if (_applicationUrl == null)
+            if (_applicationUrl != null)
+            {
+                return _applicationUrl;
+            }
+
```
blmeyers, line 335: I know this was there already, but is this caching actually useful?  We're running controllers on a per request lifecycle (as normal) so I don't think any of our scenarios will ever even hit a case where it got calculated once and gets returned here.

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
```diff
@@ -322,34 +326,35 @@ private EntityTagHeaderValue GetEtag(string etagValue)
...
-                    {
-                        urlBuilder.Port = Request.Host.Port.Value;
-                    }
-
-                    _applicationUrl = urlBuilder.Uri.ToString();
-                }
-                catch (Exception ex)
+                    Scheme = Request.Scheme,
+                    Host = Request.Host.Host,
+                    Path = Request.Path
```
blmeyers, line 342: None of this code is using the reverse proxy header support.  This means that in a reverse proxy scenario that the url will reflect the actual server that received the request rather than the reverse proxy.  Now since this is being used for a Location header, most of the reverse proxies have support for automatically (or by configuration) rewriting the location header.  For example using our local nginx container testing, this all worked fine because nginx rewrites by default.  However, the concern is that each reverse proxy requires potentially different configuration, and that they don't always handle the schema portion.  For example, AWS application load balancer will correctly rewrite the host portion of the location header, but in an ssl termination use case it won't actually change the schema from http to https, which can cause some confusion.  There's also the risk that someone could set up reverse proxy headers, verify everything is working with the version endpoint, and then get caught off guard by the location header not working quite right due to a misconfiguration.  I would lean towards modifying this code to use the same reverse proxy code as the other places we generate absolute urls, and not rely on the reverse proxy to fix the header value.  CC: @vimayya

blmeyers (changes requested):

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
vimayya: Yes. Would like to have reverse proxy supported in location header. There are helper methods in HttpRequestExtensions, so I am thinking it would be a small change.

blmeyers (approved):

PR #350, opened 2021-08-24
[ODS-5056] Allow suppression of in-memory caching of specific person types (Student, Staff, Parent) through a configuration setting

gmcelhanon:


Application/EdFi.Ods.Api/Container/Modules/PersistenceModule.cs
```diff
@@ -165,6 +165,33 @@ protected override void Load(ContainerBuilder builder)
...
                             return TimeSpan.FromSeconds(period);
                         }))
+                .WithParameter(
+                    new ResolvedParameter(
+                        (p, c) => p.Name.Equals("suppressStudentCache", StringComparison.InvariantCultureIgnoreCase),
+                        (p, c) =>
+                        {
+                            var configuration = c.Resolve<IConfiguration>();
+
+                            return configuration.GetValue<bool?>("Caching:PersonUniqueIdToUsi:SuppressStudentCache") ?? true;
```
blmeyers, line 175: With the boolean swap on these, is the intent for caching to be "suppressed" by default?  This seems to be saying if an implementation doesn't explicitly configure this setting that they'll default to suppressing the cache?

blmeyers (changes requested):
Small question around the defaults.

PR #354, opened 2021-09-08
[ODS-5042] Implement View script to support StudentUSIToEducationOrganizationIdThroughEdOrgAssociation

tbmeador-dlp:


tests/EdFi.Ods.Api.IntegrationTests/ManualPostgreSqlTests.cs
```diff
@@ -63,6 +63,16 @@ protected override IDbConnection BuildTestConnection()
             }
         }
 
+        [TestFixture]
+        [Explicit]
+        public class PostgreSqlStudentUsiToEducationOrganizationIdThroughEdOrgAssociationAuthViewTests : StudentUsiToEducationOrganizationIdThroughEdOrgAssociationAuthViewTests
```
blmeyers, line 68: I went and ran the full set of tests, and it looks like we finally got enough tests to uncover a different bug.  It looks like when the base infrastructure setup was done that it was missed to dispose the Connection object as part of tear down.  This is causing later PostgreSQL tests to fail due to the connection pool being exhausted when you run them all together.  Can you please add "Connection?.Dispose();" to the TearDown method in EdFi.Ods.Api.IntegrationTests.DatabaseTestFixtureBase to avoid this issue?

blmeyers (changes requested):

blmeyers (approved):

PR #367, opened 2021-09-21
[ODS-5049]Authorization failure error messages

semalaiappan:


semalaiappan:
Here , we are passing invalid school id here and previously error message of 
Authorization denied.  The claim does not have any established relationships with the requested resource 
but with new code , there is no way to differentiate between whether  claim does not have  edorg type or authorization is not available so we are closing this ticket . we will keep as it with old code "Authorization denied.  The claim does not have any established relationships with the requested resource"
image
image

PR #371, opened 2021-09-29
[ODS-5059] Implement code changes to redirect Write auth checks to use the new tuple table only

semalaiappan:


semalaiappan:
Please ignore failed MetaED Integration build and not related with this code change

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -82,6 +82,27 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
                     // Perform defensive checks against the remote possibility of SQL injection attack
                     ValidateTableNameParts(claimEndpointName, subjectEndpointName, authorizationSegment.AuthorizationPathModifier);
 
+                    var tupleBackedByNewAuthViews = new List<string>
```
AxelMarquez, line 85: A small improvement would be to use HashSet<string> to have constant search times (instead of linear), the difference would be small though since the list has few items

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -94,6 +115,12 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
                         continue;
                     }
 
+                    if (derivedAuthorizationViewName.Contains(authViewName))
```
blmeyers, line 118: I'm not following the logic for this.  We're coming up with an auth name based on subject to claim, then doing a "contains" check against it for the view name coming back from ViewNameHelper, to determine if we should rename the subject/claim endpoint names.  I'm assuming the contains is to try to work around scenarios with the AuthorizationPathModifier?  We're also modifying the subject and claim endpoint name in some scenarios, so not even sure if this logic path would get hit. This is one of the last tickets so I'd expect us to be moving away from hard coding at this point, but we seem to be doubling down on it further. We need to be writing this to correctly handle transforming the incoming subject and claim names in a clear and consistent way into the appropriate view name and endpoint names in order to drive the later logic for doing the auth.

blmeyers (changes requested):

Application/EdFi.Ods.Api/Security/Authorization/ViewNameHelper.cs
```diff
@@ -40,6 +40,11 @@ public static string GetAuthorizationViewName(string endpoint1, string endpoint2
                 return $"{endpoint1}To{endpoint2}{authorizationPathModifier}";
             }
 
+            if (endpoint1.EqualsIgnoreCase("EducationOrganizationId") && endpoint2.EqualsIgnoreCase("EducationOrganizationId"))
+            {
+                return $"{endpoint1}To{endpoint2}{authorizationPathModifier}";
+            }
+
```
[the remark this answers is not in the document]
semalaiappan: I removed it . Thanks

Application/EdFi.Ods.Api/Security/Utilities/AuthorizationViewsProvider.cs
```diff
@@ -34,7 +34,7 @@ private IList<string> LoadAuthorizationViews()
             using (var session = _sessionFactory.OpenStatelessSession())
             {
                 return session.CreateSQLQuery(
-                        "SELECT CONCAT(TABLE_SCHEMA,'.',TABLE_NAME) FROM INFORMATION_SCHEMA.VIEWS WHERE TABLE_SCHEMA = 'auth'")
+                        "SELECT CONCAT(TABLE_SCHEMA,'.',TABLE_NAME) FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'auth'")
```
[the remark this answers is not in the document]
semalaiappan: I did rename to reflect like what I did for codegen change.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
semalaiappan: @blmeyers  I updated the code now but still this is correct appoarch or not , I am not sure about it .

semalaiappan: @blmeyers I missed about scenarios with the AuthorizationPathModifier? ,So I moved logic within public  ClaimsMustBeAssociatedWith method in AuthorizationBuilder.cs

gmcelhanon (changes requested):

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -82,6 +83,13 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
                     // Perform defensive checks against the remote possibility of SQL injection attack
```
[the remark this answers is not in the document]
semalaiappan: we have ticket https://tracker.ed-fi.org/browse/ODS-5120 to  make the view names to be representative of {SourceClaim}To{TargetResource}, which means the views for Student should be "EducationOrganizationIdToStudentUSI".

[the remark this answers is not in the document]
semalaiappan: it looks to me this is needed (var claimEndpointsGroupedByName = authorizationSegment.ClaimsEndpoints.GroupBy(...))

Application/EdFi.Ods.Api/Security/Utilities/IAuthorizationTableViewsProvider.cs
```diff
@@ -0,0 +1,14 @@
+﻿// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace EdFi.Ods.Api.Security.Utilities
+{
+    public interface IAuthorizationTableViewsProvider
```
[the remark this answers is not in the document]
semalaiappan: I fixed it

Application/EdFi.Ods.Tests/EdFi.Ods.Security/Authorization/AuthorizationSegmentSqlProviderTests.cs
```diff
@@ -170,7 +171,7 @@ public void Should_generate_valid_sql_and_parameters()
 )
```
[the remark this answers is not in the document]
semalaiappan: yes you are correct .

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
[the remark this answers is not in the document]
semalaiappan: @gmcelhanon  I do understand now . Actually StaffUSIToEducationOrganization view was missed to handle it ,so I fixed with latest changes in code and test methods . ViewHelper file is updated but still needs to handle old legacy code for  Authorization filters in the old way . AuthorizationSegmentsToFiltersConverter.cs  line 42 string filterName = ViewNameHelper.GetAuthorizationViewName(

example for ParentUSIToSchoolId filter and view is auth.ParentUSIToEducationOrganizationId

Application/EdFi.Ods.Tests/EdFi.Ods.Security/AuthorizationStrategies/Relationships/AuthorizationSegmentsToFilterConverterTests.cs
```diff
@@ -264,11 +264,6 @@ public void Should_return_filters_for_each_associated_EdOrg_type()
                 _actualFilters[0].ClaimEndpointName.ShouldBe("LocalEducationAgencyId");
```
gmcelhanon, line 264: I'm not going to review these tests until the implementation stabilizes. I'll revisit these when the rest of the implementation is stable.

gmcelhanon (changes requested):

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -46,6 +47,22 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+            authorizationSegments.ToList().ForEach(eachSegment =>
+            {
+                eachSegment.ClaimsEndpoints.ToList().ForEach(eachClaimEndpoint =>
+                {
+                    if (EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier(eachClaimEndpoint.Name))
+                    {
+                        eachClaimEndpoint.Name = "EducationOrganizationId";
+                    }
+                    
+                    if (!PersonEntitySpecification.IsPersonIdentifier(eachSegment.SubjectEndpoint.Name))
```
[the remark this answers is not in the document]
semalaiappan: This story is very complex one for me .  What I know is I need to set every ClaimEndpoint which of EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier exists  to "EducationOrganizationId" and set every non person subjectendpoint to "EducationOrganizationId" so that it will be like EducationOrganizationIdToEducationOrganizationId tuple table to make query for authorization

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,83 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
-                    .GroupBy(ep => ep.Name)
-                    .Select(g => g)
-                    .ToList();
+                string claimEndpointName = authorizationSegment.ClaimsEndpoints.FirstOrDefault().Name;
 
-                var segmentExpressions = new List<string>();
+                if (EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier(claimEndpointName))
+                {
+                    claimEndpointName = "EducationOrganizationId";
+                }
```
gmcelhanon, line 59: Let's go ahead and close a possible exceptional condition that we're not supporting right now by adding an `else` clause and a `throw new NotSupportedException($"Claim endpoint name of '{claimEndpointName}' is not yet supported for authorization.");`.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,83 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+
+                segmentStatements.Add(CreateSegmentExpression(ref parameterIndex));
+
+                string CreateSegmentExpression(ref int index)
+                {
+                    return string.Format(
+                        StatementTemplate,
+                        GetMultiValueCriteriaExpression(authorizationSegment.ClaimsEndpoints.ToList(), parameters, ref index),
+                        subjectEndpointName.ContainsIgnoreCase("EducationOrganizationId")
+                        ? "TargetEducationOrganizationId" : subjectEndpointName,
```
gmcelhanon, line 110: If this line is to be broken like this, `?` needs to be indented and `:` needs to be on the next line, aligned with the `?` please.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,83 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+                string claimEndpointName = authorizationSegment.ClaimsEndpoints.FirstOrDefault().Name;
 
-                var segmentExpressions = new List<string>();
+                if (EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier(claimEndpointName))
+                {
+                    claimEndpointName = "EducationOrganizationId";
+                }
 
-                var unsupportedAuthorizationViews = new List<string>();
+                string subjectEndpointName = authorizationSegment.SubjectEndpoint.Name;
```
gmcelhanon, line 61: Continuing with my previous comment on immutability... please, as much as possible, try to assign variables once. Rewrite this block as follows:

```
string subjectEndpointName = EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier(authorizationSegment.SubjectEndpoint.Name)
    ? "EducationOrganizationId"
    : authorizationSegment.SubjectEndpoint.Name;
```
And once it is set... leave the value alone.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,83 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+                }
+
+                segmentStatements.Add(CreateSegmentExpression(ref parameterIndex));
+
+                string CreateSegmentExpression(ref int index)
+                {
+                    return string.Format(
+                        StatementTemplate,
+                        GetMultiValueCriteriaExpression(authorizationSegment.ClaimsEndpoints.ToList(), parameters, ref index),
+                        subjectEndpointName.ContainsIgnoreCase("EducationOrganizationId")
```
gmcelhanon, line 109: I'm asking questions here... not telling you to change anything yet.

Is `ContainsIgnoreCase` the right function here? True question. First, why is it "Contains"? Are you expecting to be handling a value with a prefix or suffix applied? Next, shouldn't this logic apply to any be any EdOrgId (or EdOrg-derived Id)? If the subject endpoint is an EdOrg (e.g. SchoolId), then shouldn't the value be supplied to the template format as "TargetEducationOrganizationId" too? This doesn't seem cohesive, but I'm only looking at this in GitHub right now.

One other thought: if this logic changes, let's try to avoid checking the subject endpoint multiple times to see if it is an EdOrg. We've already done it above and perhaps you save the results of that into a `bool` parameter and use that instead of performing multiple checks. This particular code is going to be running _a lot_.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,83 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+            }
+
+            if (unsupportedAuthorizationViews.Any())
+            {
+                _logger.Debug("Unable to authorize resource item because none of the following authorization views exist: "
+                    + $"'{string.Join("', '", unsupportedAuthorizationViews)}'");
+
+                var edOrgTypeValues = string.Join("', '", edOrgTypes.Distinct().OrderBy(x => x));
+
+                throw new EdFiSecurityException(
```
gmcelhanon, line 123: This error is based on the idea that if a view doesn't exist for a concrete-to-concrete authorization segment that the authorization isn't supported. This is no longer the case. The only time that this condition will hit is if there is an authorization path modifier that comes from the security metadata and there is no view that supports that. This condition no longer has anything to do with the EdOrg types in the claims (technically, all EdOrg sub types are expected to be supported, but this depends on the data written into the EdOrgIdToEdOrgId table) -- it's now entirely about the authorization segment.

Change the error message to reflect that reality with this new view strategy.

Application/EdFi.Ods.Api/Security/Authorization/FilterNameHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Gets the convention-based authorization filter name (without the schema), constructed by sorting the specified
+        /// endpoint names alphabetically and adding the specified modifier as a suffix (if present).
+        /// </summary>
+        /// <param name="endpoint1">The name of one endpoint represented in the view.</param>
+        /// <param name="endpoint2">The name of the other endpoint represented in the view.</param>
+        /// <param name="authorizationPathModifier">The path modifier for authorization, if applicable, indicating an alternative mechanism for
+        /// establishing a relationship between the two supplied endpoints.</param>
+        /// <returns>The authorization view name (without the schema).</returns>
+        public static string GetAuthorizationFilterName(string endpoint1, string endpoint2, string authorizationPathModifier)
```
gmcelhanon, line 24: Good.

Application/EdFi.Ods.Api/Security/Authorization/ViewNameHelper.cs
```diff
@@ -18,48 +19,28 @@ public static class ViewNameHelper
...
-                return $"{endpoint1}To{endpoint2}{authorizationPathModifier}";
-            }
-
-            if (string.Compare(endpoint1, endpoint2, StringComparison.InvariantCultureIgnoreCase) < 0)
-            {
-                return $"{endpoint1}To{endpoint2}{authorizationPathModifier}";
-            }
-
-            return $"{endpoint2}To{endpoint1}{authorizationPathModifier}";
+            return $"{subject}To{claim}{authorizationPathModifier}";
```
gmcelhanon, line 29: Just a side note here: This will need to be reversed when we clean up the views (so that it is _claim-to-subject_).

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
semalaiappan: It should be EqaulIgnoreCase instead of ContainsIgnoreCase. when subjectEndpointName is "EducationOrganizationId" and then we are calling tuple table and column name is TargetEducationOrganizationId  otherwise any person type like StaffUSI or ParentUSI or StudentUSI . So that's the reason for this change

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
semalaiappan: I updated it

Application/EdFi.Ods.Tests/EdFi.Ods.Security/AuthorizationStrategies/Relationships/AuthorizationSegmentsToFilterConverterTests.cs
semalaiappan: Please review if you think implementation is stable now .

Postman Test Suite/Ed-Fi ODS-API Integration Test Suite.postman_collection.json
```diff
@@ -182,7 +182,7 @@
 									"pm.test(\"Should return the item with SchoolId of the resource is empty string\", () => {",
 									"     ",
 									"     const responseItem = pm.response.json();",
-									"     pm.expect(responseItem.message).to.include(\"Access to the resource item could not be authorized because the \\'SchoolId\\' of the resource is empty\");",
+									"     pm.expect(responseItem.message).to.include(\"Access to the resource item could not be authorized because the \\'EducationOrganizationId\\' of the resource is empty\");",
```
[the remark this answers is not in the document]
semalaiappan: I fixed it .

semalaiappan:
@gmcelhanon   I had test method around for authorizationPathModifier for positive case and negative case

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,79 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+            var edOrgTypes = new List<string>();
 
             foreach (var authorizationSegment in authorizationSegments)
             {
-                // Within each claim segment, group the values by ed org type, and combine with "OR"
-                var claimEndpointsGroupedByName = authorizationSegment.ClaimsEndpoints
-                    .GroupBy(ep => ep.Name)
-                    .Select(g => g)
-                    .ToList();
+                string claimEndpointName = EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier(authorizationSegment.ClaimsEndpoints.FirstOrDefault().Name)
```
gmcelhanon, line 54: Don't use ternary operator to replace and compact what should just be a conditional check -- it makes the code harder to understand. Use a separate `if` statement here to determine if the claim endpoint is not an EdOrg Identifier, and then throw an exception.

Then to emphasize that `claimEndpointName` is to carry a fixed value of "EducationOrganizationId", declare it as a constant if we're still here (i.e. if no exception is thrown).

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
gmcelhanon: The error can still happen, but it will be because of the authorization path modifier... not the EdOrg types in the claims.

Application/EdFi.Ods.Api/Security/Authorization/ViewNameHelper.cs
gmcelhanon: Can we reverse the order of the parameters with a refactoring of the signature on this? I'd like to see the claim first, then the subject in the signature.

Application/EdFi.Ods.Api/Security/Authorization/ViewNameHelper.cs
```diff
@@ -18,48 +19,28 @@ public static class ViewNameHelper
...
         /// </summary>
-        /// <param name="endpoint1">The name of one endpoint represented in the view.</param>
-        /// <param name="endpoint2">The name of the other endpoint represented in the view.</param>
+        /// <param name="subject">The name of subject represented in the view.</param>
+        /// <param name="claim">The name of the claim represented in the view.</param>
         /// <param name="authorizationPathModifier">The path modifier for authorization, if applicable, indicating an alternative mechanism for
         /// establishing a relationship between the two supplied endpoints.</param>
         /// <returns>The authorization view name (without the schema).</returns>
-        public static string GetFullyQualifiedAuthorizationViewName(string endpoint1, string endpoint2, string authorizationPathModifier)
+        public static string GetFullyQualifiedAuthorizationViewName(string subject, string claim, string authorizationPathModifier)
```
gmcelhanon, line 41: Same here... claim, then subject.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
semalaiappan: Fixed it

Application/EdFi.Ods.Api/Security/Authorization/ViewNameHelper.cs
semalaiappan: Fixed it

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
```diff
@@ -45,100 +46,80 @@ public AuthorizationSegmentSqlProviderBase(IAuthorizationViewsProvider authoriza
...
+            var authorizationPathModifiers = new List<string>();
 
             foreach (var authorizationSegment in authorizationSegments)
             {
-                // Within each claim segment, group the values by ed org type, and combine with "OR"
-                var claimEndpointsGroupedByName = authorizationSegment.ClaimsEndpoints
-                    .GroupBy(ep => ep.Name)
-                    .Select(g => g)
-                    .ToList();
+                if(!EducationOrganizationEntitySpecification.IsEducationOrganizationIdentifier(authorizationSegment.ClaimsEndpoints.FirstOrDefault().Name))
```
gmcelhanon, line 54: There's a few issues with how this is written:
- It's not only the first claim that we care about, it's all of them.
- I am pretty sure that, as implemented, the same claim values will be present on all segments so the defensive check should be written to check for `Any` claims that are not EdOrgs, but it should only do this on the first segment.
- The way you wrote this, if `FirstOrDefault()` actually returns a `null`, then the subsequent `.Name` will throw an unhandled null reference exception -- very undesirable. If you used the `First()` instead, you'd get a slightly less obtuse exception message from Linq, but it's still undesirable. Try to avoid writing code with these sorts of possible outcomes by adding guard checks for the error conditions and throw custom exceptions that explain the problem to poor developer who might run into it later. In this particular case, I don't think this code execution would get this far without claims assigned, but it's better to deal with it explicitly and clearly.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationSegmentSqlProviderBase.cs
gmcelhanon: I think this error message is now too specific with regards to authorization path modifiers. It also doesn't cover any scenario related to the expansion of Student/Staff/Parent USI (which is the other thing that could change in the system to possibly cause this failure).

Can you remove the identification of the authorization path modifiers specifically, and just list the authorization view(s) that were not found but were expected to exist and leave it at that?

tests/EdFi.Ods.WebApi.CompositeSpecFlowTests/CompositeResourcesCanBeDefinedThroughXmlFeature.CompositeIncludesUnflattenedReference.approved.txt
```diff
@@ -13,6 +13,25 @@ where comp_aaa.Id = :Id AND (comp_aaa.SchoolId IN (
                             SELECT fltr_aab.StudentUSI 
                             FROM EdFi.Ods.Entities.NHibernate.QueryModels.Views.auth_StudentUSIToEducationOrganizationId fltr_aab 
                             WHERE fltr_aab.SourceEducationOrganizationId IN (:LocalEducationAgencyId)))
+order by comp_aaa.EntryDate,
```
gmcelhanon, line 16: @semalaiappan What is the change that resulted in this additional query being executed for a composite? Is this bleeding in from another set of changes where the approval test update was missed? I wouldn't expect to see it on this PR.

gmcelhanon (approved):

PR #381, opened 2021-10-12
[ODS-5128][ODS-5141] StudentUSIToEducationOrganizationIdThroughEdOrgAssociation uses wrong table

tmeador-dl:
update view name to match


gmcelhanon (changes requested):
My request for changes is related to the view name, which can't be finalized anyway. Need to discuss with @vimayya how to proceed tactically here. Rename now (for consistency with current approach) and rename again later, or commit now (with inconsistent naming) and rename once later to desired final view name.

PR #384, opened 2021-10-19
[ODS-5109] Download Test Database

raghuramakrishna:


AxelMarquez:
If strict mode is disabled, any error during download/creation of the test database should cause the tests to be skipped.
The error message should surface to the build server/user.

Application/Test.Common/MsSqlDatabaseHelper.cs
```diff
@@ -38,6 +41,49 @@ public void CopyDatabase(string originalDatabaseName, string newDatabaseName)
...
+                        ConnectionStrings = @{ EdFi_Ods = '" + _odsConnectionString + @"' } 
+                    }
+                    Set-DeploymentSettings $settings
+
+                    Reset-TestPopulatedTemplateDatabase
+                ";
+
+                var info = new ProcessStartInfo("powershell", psScript)
+                {
+                    WorkingDirectory = Path.GetFullPath("../../../../../../Ed-Fi-ODS-Implementation"),
```
AxelMarquez, line 64: How about if we pass this path as an argument? so this class remains generic, in the future there might be tests with a different folder depth.

Would there be a benefit if we also add this path to `appsettings.json` so that it is configurable?

Application/Test.Common/MsSqlDatabaseHelper.cs
```diff
@@ -38,6 +41,49 @@ public void CopyDatabase(string originalDatabaseName, string newDatabaseName)
...
+                    RedirectStandardError = true,
+                    RedirectStandardOutput = true,
+                };
+
+                using var process = Process.Start(info);
+
+                process.OutputDataReceived += (sender, e) => Console.WriteLine(e.Data);
```
AxelMarquez, line 71: Console.WriteLine doesn't seem to work in tests. A more robust approach would be to use Log4net like EdFi.Ods.Utilities.Migration.Tests does.

For now I guess we can remove these events (together with the `RedirectStandardError = true, RedirectStandardOutput = true` and `process.BeginOutputReadLine(); process.BeginErrorReadLine();` )

Application/Test.Common/MsSqlDatabaseHelper.cs
```diff
@@ -38,6 +41,49 @@ public void CopyDatabase(string originalDatabaseName, string newDatabaseName)
...
+                process.BeginOutputReadLine();
+
+                process.ErrorDataReceived += (sender, e) => Console.WriteLine(e.Data);
+                process.BeginErrorReadLine();
+
+                process.WaitForExit();
+            }
+            catch
+            { 
+                return $"Couldn't download template database or restore failed";
```
AxelMarquez, line 81: This catch statement isn't executed in case of an error, please check `process.ExitCode != 0` to see if the process failed.

May I suggest throwing an Exception instead of returning an string, perhaps `throw new InvalidOperationException("Couldn't download template database or restore failed");` and catch it in the caller.

Application/Test.Common/PgSqlDatabaseHelper.cs
```diff
@@ -34,6 +39,49 @@ public void CopyDatabase(string originalDatabaseName, string newDatabaseName)
             NpgsqlConnection.ClearAllPools();
         }
 
+        public string DownloadAndRestoreDatabase()
```
AxelMarquez, line 42: Same comments from _MsSqlDatabaseHelper.cs_ apply here

AxelMarquez (changes requested):

PR #386, opened 2021-10-27
ODS-5155 - Update trigger on EdFiOrganization to not abort transaction

bptillman:
With the trigger being an instead of trigger, when you tried to delete an EducationOrganization and the trigger failed due to a FK constraint, the trigger aborted the transaction as well, so then in the API, when it tried to roll it back, the transaction was now gone and would throw a different error instead of the intended HTTP 409 to inform the user that there was a dependency issue. With the trigger now having `SET XACT_ABORT_ OFF`, this means it will not abort and rollback the transaction its a part of, and the ODS API can handle that instead.


bptillman:
Closing this as the initially reported issue is addressed as part of changes in ODS-4817 #389

PR #388, opened 2021-10-28
[ODS-5161] Update TokenInfo provider to use auth.EducationOrganizationIdToEducationOrganizationId tuple table instead of EducationOrganizationIdentifiers auth view

semalaiappan:


bptillman (approved):

gmcelhanon (changes requested):
@semalaiappan Please review the "Implementation Notes" that I added to the ticket for [ODS-5161](https://tracker.ed-fi.org/browse/ODS-5161), and reimplement this according to that documentation. Be sure to review the dotnet fiddle that I added there as well -- you should be able to reuse that code almost in its entirety for this ticket.

Also I think we decided to just use joins here and not include the Discriminator in the `edfi.EducationOrganizationIdToEducationOrganizationId` table. Do not modify the EdOrgToEdOrgId auth triggers (which are now _generated_). We will use that approach going forward.

Eliminate all other artifacts that you can now, including the `EducationOrganizationIdentifiers` class, if possible.

tbmeador-dlp (approved):

PR #394, opened 2021-11-11
ODS-5070 - Handle TargetInvocationException so correct validation results gets returned from api

bptillman:
When sending an invalid calendarTypeDescriptoron the Calendar POST, a TargetInvocationException was being thrown, which then prevented the validation results from being returned by the API. Now we will catch this and then log the correct validation error.

Additionally, the Trace.Errors were swapped out for using log4net so that these types of messages could be seen in the WebApiLog.

### DIFF
--- a/Application/EdFi.Ods.Common/Extensions/ValidatorExtensions.cs
+++ b/Application/EdFi.Ods.Common/Extensions/ValidatorExtensions.cs
@@ -6,12 +6,15 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
-using System.Diagnostics;
+using System.Reflection;
+using log4net;
 
 namespace EdFi.Ods.Common.Extensions
 {
     public static class ObjectValidatorExtensions
     {
+        private static readonly ILog _logger = LogManager.GetLogger(typeof(ObjectValidatorExtensions));
+
         public static ICollection<ValidationResult> ValidateObject(this IEnumerable<IObjectValidator> validators, object @object)
         {
             var result = new List<ValidationResult>();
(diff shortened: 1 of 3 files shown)


gmcelhanon (changes requested):
This is one of those that it would be easy and good to add a simple unit test for coverage. I reproduced the problem in Linqpad, but here's the basic approach.

Declare a test class with a validation attribute that throws an exception when its value is retrieved, like so:
```csharp
public class ClassWithValidationThrowingException
{
	private string _value;
	
	[RequiredWithNonDefault]
	public string Value
	{
		get => throw new Exception("Test validation failure message.");
		set => _value = value;
	}
}
```

The unit test body should execute the validation as follows (add appropriate assertions):
```csharp
    var p = new ClassWithValidationThrowingException { Value = "Abc" };
	
    var validators = new IEntityValidator[] { new DataAnnotationsEntityValidator() };
    var results = validators.ValidateObject(p);

    // Assert that validation failed and that the validation message is the exception text from the test class.
```
Follow conventions and put the class in this folder:
image

We have 2 main styles of unit tests. A more traditional barebones NUnit test fixture with individual wholly contained test methods named like `When_doing_this_thing_should_see_this_result`.  Alternatively, you can inherit from `TestFixtureBase` and override Arrange/Assert/Act methods to write the test. This is generally reserved for more involved unit tests with dependencies and involves Acting once in the fixture, saving the results for subsequent inspection in individual assertions declared as test methods.

bptillman:
> ```cs
> 	private string _value;
> 	
> 	[RequiredWithNonDefault]
> 	public string Value
> 	{
> 		get => throw new Exception("Test validation failure message.");
> 		set => _value = value;
> 	}
> ```

A new test has been added per your guidelines here.

gmcelhanon (approved):

semalaiappan (approved):

PR #395, opened 2021-11-12
[ODS-5127] Data Standard Metadata Artifacts cleanup 

semalaiappan:


bptillman (approved):
All things look to be good with these files being removed.

PR #396, opened 2021-11-13
[ODS-4944] Session Name case in studentSchoolAttendanceEvnts is not enforced in POST, but breaks GET

gmcelhanon:


bptillman (approved):
This looks good for merging. I tested this before and after and it now works correctly compared to before with the bad response.

PR #397, opened 2021-11-14
[ODS-5179] Security visualization utility shows incorrect authorization strategy

AxelMarquez:
See my comment in ODS-5179 for the root cause explanation.

### DIFF
--- a/Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
+++ b/Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
@@ -363,11 +363,11 @@ order by
             // First process the segments into the graph
             foreach (var metadataEdge in distinctMetadataEdges)
             {
-                var vertex = AddAndGetVertex(resourceGraph, metadataEdge.ClaimName);
+                var vertex = GetOrAddVertex(resourceGraph, metadataEdge.ClaimName);
 
                 if (!string.IsNullOrEmpty(metadataEdge.ParentClaimName))
                 {
-                    var parentVertex = AddAndGetVertex(resourceGraph, metadataEdge.ParentClaimName);
+                    var parentVertex = GetOrAddVertex(resourceGraph, metadataEdge.ParentClaimName);
                     resourceGraph.AddEdge(new Edge<Resource>(parentVertex, vertex));
                 }
             }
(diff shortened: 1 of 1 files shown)


Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
```diff
@@ -546,6 +543,18 @@ private static List<Resource> GetRootNodes(AdjacencyGraph<Resource, Edge<Resourc
             return subgraph;
         }
 
+        private static Resource AddAndGetVertex(AdjacencyGraph<Resource, Edge<Resource>> resourceGraph, string claimName)
```
gmcelhanon, line 546: Let's follow .NET naming conventions on this one... like ConcurrentDictionary's [GetOrAdd](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-5.0) method since it has the same semantics.

Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
```diff
@@ -546,6 +543,18 @@ private static List<Resource> GetRootNodes(AdjacencyGraph<Resource, Edge<Resourc
...
+        {
+            var vertex = resourceGraph.Vertices.SingleOrDefault(v => v.Name == claimName);
+            if (vertex is null)
```
gmcelhanon, line 549: Add blank lines between blocks (between `var vertex` and `if (vertext is null)`.

gmcelhanon (commented):
A couple very minor changes requested.

Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
AxelMarquez: Thank you; done!

Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
AxelMarquez: Thank you; done!

gmcelhanon (approved):

PR #407, opened 2021-12-15
ODS-5186 - Update Dependencies to include Updates operation and ensure resources are output in correct operational order

bptillman:
Previously everything was output in the correct order, but with only a "Create" operation and no "Updates". Now this fixes it so that the graph results will include both possible operations, and then when needed, split out "Create" and "Update" based on any resources that have to be created before an update can be performed. With these changes the following order of resources should be observed in the output from this endpoint:

- Student Create / Parent Create
- StudentSchoolAssociation Create/Update
- StudentParentAssociation Create/Update
- Student Update / Parent Update
- StaffEducationOrganizationEmploymentAssociations Create/Update
- StaffEducationOrganizationAssignmentAssociations Create/Update

For additional information about this endpoint, please [see this Tech Docs page](https://techdocs.ed-fi.org/display/ODSAPIS3V53/Resource+Dependency+Order).

### DIFF
--- a/Application/EdFi.Ods.Api/Models/GraphML/ResourceLoadOrder.cs
+++ b/Application/EdFi.Ods.Api/Models/GraphML/ResourceLoadOrder.cs
@@ -13,9 +13,10 @@ namespace EdFi.Ods.Api.Models.GraphML
 
         public int Order { get; set; }
 
-        public IReadOnlyList<string> Operations
+        public IList<string> Operations { get; set; } = new List<string>
         {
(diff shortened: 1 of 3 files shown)


gmcelhanon (approved):

vishalz-cognologix:
Not sure if this is working still... I guess this might still be open. Just checked the API, it only lists Create order!
https://api.ed-fi.org/v5.3/api/metadata/data/v3/dependencies

PR #409, opened 2021-12-18
[EDFI-1067] Update the ODS/API to NET 6.0

jamessmckay:
* updated the project files to support net 6.0
* fixed the resources being generated to remove the compile error
* added global.json files to the solutions that require it

Notes
* Must install the net 6.0 sdk before proceeding
* There is a required change to codegen. This needs to be run first manually before running initdev. (run initdev -nocodegen after to build otherwise) until the image is built in team city.
* added global.json files, this ensures the application is built correctly with the right framework. May want to spin off a ticket to remove the checks in the powershell scripts.

Remaining Work
* TeamCity needs to be updated with the net 6.0 sdk
* CodeGen Package needs to be created and solution updated with that package
* CodeGen Approval Tests need to be updated
* EdFi.Common needs a new package created and solution updated
* EdFi.Admin.DataAccess needs a new package created and solution updated
* EdFi.Security.DataAccess needs a new package created and solution updated

Items not addressed
* Sandbox Admin Identity package was not updated due to the change in the model (the current model works fine with net 6.0)
* PostgreSQL packages not updated due to the unknown repercussion with Sandbox Admin
*  Fluent Validation package was not updated. This requires a migration from 8.x to 10.x and there are breaking changes, suggestion here is to spin off a ticket and do the necessary update. This will require a codegen change too.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,4 +13,4 @@ profile to this file with their first pull request.
 * **[Jean-Francois Guertin, EdWire](https://github.com/frenchyjef)**
 * **[Mansoor Khalid, EdWire](https://github.com/Mansoor-Khalid)**
 * **[Google Cloud](https://cloud.google.com/) / [Innive](https://innive.com/)**
-* **[Jim McKay, Instructure](https://github.com/jamessmckay)**
+* **[Jim McKay](https://github.com/jamessmckay)**
--- a/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/Resources.mustache
+++ b/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/Resources.mustache
@@ -579,7 +579,6 @@ namespace {{ResourceClassesNamespace}}
         /// </returns>
         public override bool Equals(object obj)
         {
-            #pragma warning disable 472
             var compareTo = obj as {{NamespacePrefix}}I{{EntityName}};
 
             if (ReferenceEquals(this, compareTo))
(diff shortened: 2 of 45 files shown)


CONTRIBUTORS.md
```diff
@@ -13,4 +13,4 @@ profile to this file with their first pull request.
 * **[Jean-Francois Guertin, EdWire](https://github.com/frenchyjef)**
 * **[Mansoor Khalid, EdWire](https://github.com/Mansoor-Khalid)**
 * **[Google Cloud](https://cloud.google.com/) / [Innive](https://innive.com/)**
-* **[Jim McKay, Instructure](https://github.com/jamessmckay)**
+* **[Jim McKay](https://github.com/jamessmckay)**
```
stephenfuqua, line 16: (☞ﾟヮﾟ)☞

Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/Resources.mustache
```diff
@@ -598,49 +597,82 @@ namespace {{ResourceClassesNamespace}}
             {{#Usi}}
             {{#Property}}
             // Property
+            {{#IsNullable}}
             if ((this as {{NamespacePrefix}}I{{ClassName}}).{{PropertyName}} == null
                 || !(this as {{NamespacePrefix}}I{{ClassName}}).{{PropertyName}}.Equals(compareTo.{{PropertyName}}))
                 return false;
+            {{/IsNullable}}
+            {{^IsNullable}}
+            if ((this as {{NamespacePrefix}}I{{ClassName}}).{{PropertyName}}.Equals(compareTo.{{PropertyName}}))
```
stephenfuqua, line 606: I think I understand what the template is doing. The question is, why make this change now? Is this solving some known bug?

jamessmckay: the compiler does not respect the pragma statement anymore, so i needed to fix code gen so the application would compile

PR #413, opened 2021-12-31
[ODS-5221] Add Postman tests for Record Ownership feature

semalaiappan:
Use case for Record Ownership Testing using School and StudentSectionAssociation  resource

   --Client 1 should be able to full CRUD their School 1.
  --Resource -Client 1 has full CRUD - School 1
  --Client 2 should be able to full CRUD School 2.
  -- Resource -Client 2 has full CRUD - School 2
  --Assuming Client 2 is assigned the extra ownership token, it should also be able to full CRUD School 1. 
  --But Client 1, without the assignment to Ownership Token 2, should not be able to CRUD School 2.

### DIFF
--- a/Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
+++ b/Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
@@ -217,10 +217,7 @@
 											"listen": "prerequest",
 											"script": {
 												"exec": [
-													"const scenarioId = pm.environment.get('scenarioId');",
-													"const uuid = require('uuid');",
-													"function newGuid() { return uuid.v4().toString().replace(/[^a-zA-Z0-9 ]/g,\"\"); }",
-													"pm.environment.set('supplied:'+scenarioId+':nameOfInstitution',newGuid());"
+													""
 												],
 												"type": "text/javascript"
 											}
@@ -231,6 +228,26 @@
 												"exec": [
 													"pm.test(\"Status code is 204\", () => {",
 													"    pm.expect(pm.response.code).to.equal(204);",
+													"});",
+													"",
+													"let SchoolsGetRequest = {",
+													"    url: pm.environment.get(\"ApiBaseUrl\") + \"/data/v3/ed-fi/schools/\"+ pm.environment.get(\"known:School1:schoolGuid\"),",
+													"    method: 'GET',",
+													"    header: 'Authorization:bearer ' + pm.environment.get(\"AccessToken_255901-A\"),",
+													"    body: {}",
+													"};",
+													"",
+													"const scenarioId = pm.environment.get('scenarioId');",
+													"pm.test('nameOfInstitution contains substring of \"updated\"  ', () => {",
+													"",
+													"    pm.sendRequest(SchoolsGetRequest, function (err, Schools) {",
+													"        if (err) {",
+													"            console.log(err);",
+													"        } else {",
+													"            const responseItem = Schools.json();",
+													"                pm.expect(responseItem.nameOfInstitution).to.equal('Updated '+pm.environment.get('supplied:'+scenarioId+':nameOfInstitution').toString());",
+													"        }",
+													"    });",
 													"});"
 												],
 												"type": "text/javascript"
@@ -259,7 +276,7 @@
 										],
 										"body": {
 											"mode": "raw",
-											"raw": "{\r\n    \"educationOrganizationCategories\": [\r\n    {\r\n      \"educationOrganizationCategoryDescriptor\":  \"{{supplied:{{scenarioId}}:educationOrganizationCategoryDescriptor}}\"\r\n    }\r\n  ],\r\n   \"gradeLevels\": [\r\n    {\r\n      \"gradeLevelDescriptor\":  \"{{supplied:{{scenarioId}}:gradeLevelDescriptor}}\"\r\n    }\r\n  ],\r\n \r\n  \"nameOfInstitution\":  \"{{supplied:{{scenarioId}}:nameOfInstitution}}\",\r\n  \"schoolCategories\": [\r\n    {\r\n      \"schoolCategoryDescriptor\": \"{{supplied:{{scenarioId}}:schoolCategoryDescriptor}}\"\r\n    }\r\n  ],\r\n  \"shortNameOfInstitution\":  \"{{supplied:{{scenarioId}}:nameOfInstitution}}\",\r\n  \"schoolTypeDescriptor\": \"{{supplied:{{scenarioId}}:schoolTypeDescriptor}}\"\r\n}"
+											"raw": "{\r\n    \"educationOrganizationCategories\": [\r\n    {\r\n      \"educationOrganizationCategoryDescriptor\":  \"{{supplied:{{scenarioId}}:educationOrganizationCategoryDescriptor}}\"\r\n    }\r\n  ],\r\n   \"gradeLevels\": [\r\n    {\r\n      \"gradeLevelDescriptor\":  \"{{supplied:{{scenarioId}}:gradeLevelDescriptor}}\"\r\n    }\r\n  ],\r\n \r\n  \"nameOfInstitution\":  \"Updated {{supplied:{{scenarioId}}:nameOfInstitution}}\",\r\n  \"schoolCategories\": [\r\n    {\r\n      \"schoolCategoryDescriptor\": \"{{supplied:{{scenarioId}}:schoolCategoryDescriptor}}\"\r\n    }\r\n  ],\r\n  \"shortNameOfInstitution\":  \"{{supplied:{{scenarioId}}:nameOfInstitution}}\",\r\n  \"schoolTypeDescriptor\": \"{{supplied:{{scenarioId}}:schoolTypeDescriptor}}\"\r\n}"
 										},
 										"url": {
 											"raw": "{{ApiBaseUrl}}/data/v3/ed-fi/schools/{{known:School1:schoolGuid}}",
@@ -428,20 +445,12 @@
 									"response": []
 								},
 								{
-									"name": "Create  School",
+									"name": "Update  School",
 									"event": [
 										{
 											"listen": "prerequest",
 											"script": {
 												"exec": [
-													"const scenarioId = pm.environment.get('scenarioId');",
-													"",
-													"pm.environment.set('supplied:'+scenarioId+':nameOfInstitution', 'SCH-9011');",
-													"pm.environment.set('supplied:'+scenarioId+':educationOrganizationCategoryDescriptor', \"uri://ed-fi.org/EducationOrganizationCategoryDescriptor#Local Education Agency\");",
-													"pm.environment.set('supplied:'+scenarioId+':gradeLevelDescriptor', \"uri://ed-fi.org/GradeLevelDescriptor#Adult Education\");",
-													"pm.environment.set('supplied:'+scenarioId+':schoolCategoryDescriptor', \"uri://ed-fi.org/SchoolCategoryDescriptor#Adult School\");",
-													"pm.environment.set('supplied:'+scenarioId+':schoolTypeDescriptor', \"uri://ed-fi.org/SchoolTypeDescriptor#Regular\");",
-													"",
 													""
 												],
 												"type": "text/javascript"
@@ -455,7 +464,26 @@
 													"    pm.expect(pm.response.code).to.equal(200);",
 													"});",
 													"pm.environment.set('known:School1:schoolGuid',pm.response.headers.one('Location').value.split(\"/\").pop());",
-													""
+													"",
+													"let SchoolsGetRequest = {",
+													"    url: pm.environment.get(\"ApiBaseUrl\") + \"/data/v3/ed-fi/schools/\"+ pm.environment.get(\"known:School1:schoolGuid\"),",
+													"    method: 'GET',",
+													"    header: 'Authorization:bearer ' + pm.environment.get(\"AccessToken_255901-B\"),",
+													"    body: {}",
+													"};",
+													"",
+													"const scenarioId = pm.environment.get('scenarioId');",
+													"pm.test('nameOfInstitution contains substring of \"updated\"  ', () => {",
+													"",
+													"    pm.sendRequest(SchoolsGetRequest, function (err, Schools) {",
+													"        if (err) {",
+													"            console.log(err);",
+													"        } else {",
+													"            const responseItem = Schools.json();",
+													"                pm.expect(responseItem.nameOfInstitution).to.equal('Updated '+pm.environment.get('supplied:'+scenarioId+':nameOfInstitution').toString());",
+													"        }",
+													"    });",
+													"});"
 												],
 												"type": "text/javascript"
 											}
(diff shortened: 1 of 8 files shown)


Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+
+SELECT @OwnershipToken_OwnershipTokenId = OwnershipTokenId FROM [dbo].[OwnershipTokens] where [Description] in ('OwnershipToken_One')
+INSERT INTO [dbo].[ApiClientOwnershipTokens]([ApiClient_ApiClientId],[OwnershipToken_OwnershipTokenId]) VALUES (@ApiClient_ApiClientId,@OwnershipToken_OwnershipTokenId)
+
+UPDATE [dbo].[ApiClients]
+   SET CreatorOwnershipTokenId_OwnershipTokenId = @OwnershipToken_OwnershipTokenId
+ WHERE ApiClientId =@ApiClient_ApiClientId
+
+
+SELECT @ApiClient_ApiClientId = ApiClientId FROM [dbo].[ApiClients] where Name in ('255901044')
```
gmcelhanon, line 26: It's best not to reuse variables like this. It increases maintenance costs and likelihood of bugs in the future. Declare new variables for each value, set them once, and then don't change them.

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+
+SELECT @OwnershipToken_OwnershipTokenId = OwnershipTokenId FROM [dbo].[OwnershipTokens] where [Description] in ('OwnershipToken_Two')
+INSERT INTO [dbo].[ApiClientOwnershipTokens]([ApiClient_ApiClientId],[OwnershipToken_OwnershipTokenId]) VALUES (@ApiClient_ApiClientId,@OwnershipToken_OwnershipTokenId)
+
+UPDATE [dbo].[ApiClients]
+   SET CreatorOwnershipTokenId_OwnershipTokenId = @OwnershipToken_OwnershipTokenId
+ WHERE ApiClientId =@ApiClient_ApiClientId
+
+
+SELECT CreatorOwnershipTokenId_OwnershipTokenId,*  FROM [dbo].[ApiClients]  where Name in ('255901044','255901001')
```
gmcelhanon, line 37: These `SELECT` statements look to be debugging/diagnostic in nature. They don't serve any functional purpose in the script, so they should be removed.

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+
+use EdFi_Admin_Test
+GO
+
+INSERT INTO [dbo].[OwnershipTokens] ([Description])     VALUES  ('OwnershipToken_One')
+INSERT INTO [dbo].[OwnershipTokens] ([Description])     VALUES  ('OwnershipToken_Two')
+
+DECLARE @ApiClient_ApiClientId INT;
+DECLARE @OwnershipToken_OwnershipTokenId INT;
+SELECT @ApiClient_ApiClientId = ApiClientId FROM [dbo].[ApiClients] where Name in ('255901001')
```
gmcelhanon, line 16: Are we using the "No Further Authorization Required" strategy for educationOrganizations in the security metadata for this testing, and are we granting full CRUD permissions to educationOrganizations to both of these clients? I ask that because if so, it's a bit irrelevant (even non-intuitive or misleading) if they are API clients affiliated with specific Schools, because typically that would provide relationship-based constraints on what data they are allowed modify.

Ideally, we'd have 2 clients that associated with the same LEA so that there's no concerns over general access to the target resource based on relationships. Then, the only factor that obviously needs to be considered in their 403 when trying to update or delete the resource is the Record Ownership strategy.

Set up the test scenario so that the reason for the 403 is both _distinct_ and _obvious_.

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+SELECT *  FROM [dbo].[OwnershipTokens] where [Description] in ('OwnershipToken_One','OwnershipToken_Two')
+
+USE [EdFi_Security_Test]
+Go
+
+DECLARE @AuthorizationStrategyId INT;
+
+INSERT INTO [dbo].[AuthorizationStrategies]   ([DisplayName],[AuthorizationStrategyName],[Application_ApplicationId]) VALUES ('Ownership Based','OwnershipBased', 1)
+
+SELECT @AuthorizationStrategyId=AuthorizationStrategyId from [dbo].[AuthorizationStrategies] where [AuthorizationStrategyName]='OwnershipBased'
```
gmcelhanon, line 48: As an FYI, when an insert statement operates on a table with an IDENTITY column like the EF tables do, you can use  the `SCOPE_IDENTITY()` function here to get the Id instead of writing the query.

Also, rename this variable to be clearer for the purpose it is used for -- `@OwnershipBasedAuthorizationStrategyId`.

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+SELECT CreatorOwnershipTokenId_OwnershipTokenId,*  FROM [dbo].[ApiClients]  where Name in ('255901044','255901001')
+SELECT *  FROM [dbo].[ApiClientOwnershipTokens] where ApiClient_ApiClientId in ( SELECT ApiClientId  FROM [dbo].[ApiClients]  where Name in ('255901044','255901001'))
+SELECT *  FROM [dbo].[OwnershipTokens] where [Description] in ('OwnershipToken_One','OwnershipToken_Two')
+
+USE [EdFi_Security_Test]
+Go
+
+DECLARE @AuthorizationStrategyId INT;
+
+INSERT INTO [dbo].[AuthorizationStrategies]   ([DisplayName],[AuthorizationStrategyName],[Application_ApplicationId]) VALUES ('Ownership Based','OwnershipBased', 1)
```
gmcelhanon, line 46: Shouldn't we have a place elsewhere in the change scripts for the Ownership Based authorization feature that already adds this record?

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+
+  INSERT INTO  [dbo].[ResourceClaimActionAuthorizationStrategies]
+ ([ResourceClaimActionId],[AuthorizationStrategyId]) 
+ SELECT ResourceClaimActionId,@AuthorizationStrategyId  FROM [dbo].ResourceClaimActions rca 
+INNER JOIN  [dbo].ResourceClaims rc  on rca.ResourceClaimId = rc.ResourceClaimId
+AND ClaimName='http://ed-fi.org/ods/identity/claims/studentSectionAssociation'
+INNER JOIN  [dbo].Actions a  on a.ActionId= rca.ActionId 
+AND a.ActionName in ('Read','Update','Delete')
+
+SELECT *  FROM [dbo].ResourceClaimActionAuthorizationStrategies
```
gmcelhanon, line 76: This looks to be diagnostic in nature and shouldn't be in this script.

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
```diff
@@ -0,0 +1,82 @@
...
+
+DECLARE @AuthorizationStrategyId INT;
+
+INSERT INTO [dbo].[AuthorizationStrategies]   ([DisplayName],[AuthorizationStrategyName],[Application_ApplicationId]) VALUES ('Ownership Based','OwnershipBased', 1)
+
+SELECT @AuthorizationStrategyId=AuthorizationStrategyId from [dbo].[AuthorizationStrategies] where [AuthorizationStrategyName]='OwnershipBased'
+
+
+ --education organization resource claims for read update and delete
+ INSERT INTO  [dbo].[ResourceClaimActionAuthorizationStrategies]
```
gmcelhanon, line 52: This script you're writing here is updating the _default_ Ed-Fi security metadata (which is not an approach generally recommended).

While it's definitely good to cover it in our tests, for full test coverage we really would need to also set up metadata with Claim Set specific overrides (to the API clients' assigned Claim Set) that apply the Ownership-based authorization strategy.

The tricky part for ultimately automating this is that we'll need to restart the API process in between test scenarios so that the security metadata is refreshed by the API.

gmcelhanon (changes requested):
I haven't yet reviewed the Postman tests... but there's enough here on the SQL script I wanted to get it back to you so you can start working on them.

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
```diff
@@ -0,0 +1,3388 @@
...
+												"ed-fi",
+												"schools"
+											]
+										},
+										"description": "This api post method adds new academicWeeks for particular school .\nThis test method will throw WeekIdentifier is required error when WeekIdentifier is not passed"
+									},
+									"response": []
+								},
+								{
+									"name": "Create  School",
```
gmcelhanon: Rename this to Update School (to emphasize that the previous step actually created the School that is now being updated).

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
```diff
@@ -0,0 +1,3388 @@
...
+										"method": "POST",
+										"header": [
+											{
+												"key": "Content-Type",
+												"name": "Content-Type",
+												"type": "text",
+												"value": "application/json"
+											}
+										],
+										"body": {
```
gmcelhanon: We need to make sure that an update actually occurs. If nothing changes, NHibernate won't attempt to update at all. I suggest changing the body so that the nameOfInstitution is set to something like `"Updated {{supplied:{{scenarioId}}:nameOfInstitution}}"`.

If you do this, you'll need to update the assertions on the next step (GET), though I really wonder what those assertions are actually adding to _this_ test. This test should be focused on the response codes (did we get an error or not?). We are not really concerned here with the mechanical aspects of whether the item's correct name, category, grade level, school category, etc. are being returned, and that just creates more code to maintain without adding value through its verification.

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
```diff
@@ -0,0 +1,3388 @@
...
+												"ed-fi",
+												"schools",
+												"{{known:School1:schoolGuid}}"
+											]
+										}
+									},
+									"response": []
+								},
+								{
+									"name": "Update  School",
```
gmcelhanon: Rename this to "Update (by Id)". Same comment here with regards to ensuring the body has changed.

gmcelhanon (changes requested):
The Postman tests look pretty good. I have a left a few comments on the first few requests that pretty much apply throughout all the tests, but fundamentally they look like they cover what is needed, so it's a pretty good first pass.

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: Do you mean that under this folder where adding 'OwnershipBased' AuthorizationStrategies insert script exists here Ed-Fi-ODS\Application\EdFi.Ods.Standard\Artifacts\MsSql\Structure\Ods\Changes\

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: Fixed it .

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: removed it

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: 2 clients that associated with the same LEA now

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: i updated it

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: removed it

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
```diff
@@ -0,0 +1,3388 @@
...
+												],
+												"type": "text/javascript"
+											}
+										},
+										{
+											"listen": "test",
+											"script": {
+												"exec": [
+													"pm.test(\"Status code is 200 or 201\", () => {",
+													"    pm.expect(pm.response.code).to.be.oneOf([200, 201]);",
```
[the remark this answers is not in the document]
semalaiappan: I updated it now

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
```diff
@@ -0,0 +1,3388 @@
...
```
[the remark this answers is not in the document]
semalaiappan: I updated it now

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
semalaiappan: I updated it now

Postman Test Suite/Record Ownership Test/Ed-Fi ODS-API Record Ownership Test Suite.postman_collection.json
semalaiappan: I updated it now

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
gmcelhanon: @vimayya The security metadata for the OwnershipBased authorization strategy should be added by the feature (right?), so does that mean this `INSERT` statement should be in a script located at the following path in the Ed-Fi-ODS repository? _/Artifacts/MsSql/Data/Security/{featureName}_

Or should it be in the implementation repository (because the Ownership-based authorization feature project is in the Implementation repository)?

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
gmcelhanon: @semalaiappan Do we have coverage here for both "default" security metadata with the authorization strategy, and claim set specific override metadata that uses it? I believe that we've got unit tests around that resolution, so it's probably not critical, but we ended up with the authorization strategy being included in the "default" metadata, right?

gmcelhanon (approved):

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
vimayya: Current auth strategy creation logic is in code gen https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/main/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Mustache/ResourceClaimMetadata.mustache#L43 If there is no easy way include it only when the feature is on, we should probably keep it together with the rest in codegen.

gmcelhanon: So there's no feature/extension-based sub-folder checking done for the Security database, as is done for extensions on the ODS?

vimayya: feature based sub-folder checking may work for the Security database, but we haven't done that so far.

semalaiappan: codegen will not  know OwnershipBasedAuthorization feature is on or not at the time of generating it .  I will have insert script in codegen

Postman Test Suite/Record Ownership Test/Record Ownership Manual Setup Scripts.sql
semalaiappan: I am updating in ResourceClaimActionAuthorizationStrategies and ResourceClaimActions table for this . So it's default in Security Database .
I am not using based on any  Claimset  here.  I am not sure how to load data for Claimset Override metadata for this use case to  test whether this postman test coverage covered or not

PR #424, opened 2022-01-31
ODS-5242 - Update postman test request data to fix tests

bptillman:
Tests were failing for two reasons:
1. A null reference exception on missing data not being in the requests. Adding the new needed information got these tests passing.
2. A routing precedence change occurred in the latest update, so an ordering had to be set on the deletes endpoint to ensure they are done in the right order

Please not that other tests will still be failing on this build and will be handled by later tickets, but the postman tests should all be passing.

### DIFF
--- a/Application/EdFi.Ods.Features/Controllers/DeletesController.cs
+++ b/Application/EdFi.Ods.Features/Controllers/DeletesController.cs
@@ -19,7 +19,7 @@ namespace EdFi.Ods.Features.Controllers
     [Authorize]
     [ApiController]
     [Produces("application/json")]
-    [Route("{schema}/{resource}/deletes")]
+    [Route("{schema}/{resource}/deletes", Order = -1)]
     public class DeletesController : ControllerBase
     {
         private readonly IGetDeletedResourceIds _getDeletedResourceIdsRepository;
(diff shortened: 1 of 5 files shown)


semalaiappan:
@bptillman   It's recommended that avoid using order here https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-6.0

Application/EdFi.Ods.Features/Controllers/DeletesController.cs
```diff
@@ -19,7 +19,7 @@ namespace EdFi.Ods.Features.Controllers
     [Authorize]
     [ApiController]
     [Produces("application/json")]
-    [Route("{schema}/{resource}/deletes")]
+    [Route("{schema}/{resource}/deletes", Order = -1)]
```
semalaiappan, line 22: It's recommended that avoid using order here [docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-6.0](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-6.0)

semalaiappan (changes requested):

Application/EdFi.Ods.Features/Controllers/DeletesController.cs
bptillman: After further discussions with the team, to get tests passing so further changes can be made to get everything on NET6, this fix will stay in place for now and a subsequent ticket is going to be added for a more permanent fix to handle the [route order/precedence changes that happened as a part of the upgrade](https://github.com/dotnet/aspnetcore/issues/29594#issuecomment-767063089) . One option being looked at is moving the deletes into CodeGen to be part of each aggregate instead of the catch all we have here. But the future ticket will handle seeing if that is a viable option or if we need to handle it a different way.

AxelMarquez (approved):

PR #428, opened 2022-02-03
ODS-5260 Remove global.json files

semalaiappan:


bptillman (approved):
Confirmed no more global.json files are found and everything builds and debugs like it did before the change.

PR #429, opened 2022-02-04
[ODS-5226] Include main api area base urls in the version endpoint

ericstockdale:


ericstockdale:
Did not associate with Ticket

bptillman (changes requested):
These chagnes look good, but since we have an existing test (VersionControllerTests) for testing this controller, we should also update the test to account for these changes.

Application/EdFi.Ods.Api/Controllers/VersionController.cs
```diff
@@ -136,6 +136,41 @@ public IActionResult Get()
                                                    "xsd";
                 }
 
+                if (_apiSettings.IsFeatureEnabled(ApiFeature.ChangeQueries.GetConfigKeyName()))
+                {
+                    urlsByName["changeQueries"] = Request.RootUrl(useReverseProxyHeaders) +
+                                                  $"/changeQueries/v{ApiVersionConstants.ChangeQuery}/" +
```
bptillman, line 142: This Api Constant and the Identity one below appear to be missing, which is causing the build to fail since the solution cannot be built. Please add these, confirm the build works, and then commit and push the change so it can be re-reviewed.

bptillman (changes requested):

Application/EdFi.Ods.Api/Controllers/VersionController.cs
ericstockdale: I added the missing constants, reran initdev, reran the tests.  We should be good now.

bptillman (approved):
Changes look good to merge now.

PR #432, opened 2022-02-08
ODS-5246 - Update SDKGen to target net6

bptillman:
Also updated nuget packages to match versions they were updated to in other projects. Latest log4net is now 2.0.14, but since other projects were updated to 2.0.13, it was updated only to that version.

No updates were needed to the implementation repo since all the relevant packages pull "latest"

### DIFF
--- a/Utilities/SdkGen/EdFi.SdkGen.Console/EdFi.SdkGen.Console.csproj
+++ b/Utilities/SdkGen/EdFi.SdkGen.Console/EdFi.SdkGen.Console.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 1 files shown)


AxelMarquez (changes requested):
`Invoke-SdkGen` currently fails because it tries to find the executable in the netcoreapp3.1 folder, please rename it to net6.0 here:

https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/main/logistics/scripts/modules/TestHarness.psm1#L71

bptillman:
> `Invoke-SdkGen` currently fails because it tries to find the executable in the netcoreapp3.1 folder, please rename it to net6.0 here:
> 
> https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/main/logistics/scripts/modules/TestHarness.psm1#L71

This has been updated now to no longer look for the hard coded 3.1 path and instead look for it in a similar fashion to what we do for the test harness executable.

AxelMarquez (approved):

PR #435, opened 2022-02-11
[ODS-5270] Security visualization tool doesn't reflect AuthorizationStrategyOverrides

AxelMarquez:


Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
```diff
@@ -860,26 +871,9 @@ select new
...
-                // There was a claim set specific setting on the current resource
-                effectiveActionAndStrategy.TrySetActionGranted(inherited: false);
-
-                effectiveActionAndStrategy.TrySetAuthorizationStrategy(
-                    explicitActionAndStrategyForAction.AuthorizationStrategy,
-                    resource.Name,
-                    inherited: false,
-                    isDefault: false);
-            }
+            // Step 2: Process from top to bottom (least to most in terms of priority)
```
AxelMarquez, line 874: Had to invert the calculation of the overrides here since it was prioritizing the inherited override instead of the override that is directly associated to the resource.

bptillman (approved):

PR #436, opened 2022-02-14
[ODS-4835] Code Generation for Resource Links to some Base Aggregates Fail

gmcelhanon:
Removing unused references to base property mappings to fix issues with link generation for GeneralStudentProgramAssociation links.


bptillman (changes requested):

gmcelhanon (commented):
As discussed on Slack, the unused methods have also been removed through an additional commit by @bptillman .

tests/EdFi.Ods.WebApi.CompositeSpecFlowTests/Steps.cs
```diff
@@ -114,7 +114,10 @@ public async Task GivenTheSubjectOfTheRequestIsAStudentSchoolAssociation()
         {
             const string query = @"
                     SELECT  TOP 1 sa.Id
-                    FROM    edfi.StudentSchoolAssociation sa ";
+                    FROM    edfi.StudentSchoolAssociation sa
+                    INNER JOIN edfi.School sc 
+                            ON sa.SchoolId = sc.SchoolId
+                    WHERE sc.LocalEducationAgencyId IS NOT NULL";
```
gmcelhanon, line 120: @bptillman I was thinking we would tap into the EducationOrganizationIdToEducationOrganizationId authorization view to identify a StudentSchoolAssociation that is accessible to the test client here. With the `IS NOT NULL` check that you've added here, It is still conceivable that we could end up with the same problem if another school was created for testing purposes like the one we saw in the test ODS for SchoolId = 7 except with the LocalEducationAgencyId set.

What do you think about making this a more durable query -- something like the following:

```sql
SELECT  TOP 1 ssa.Id
FROM    edfi.StudentSchoolAssociation ssa
        INNER JOIN auth.EducationOrganizationIdToEducationOrganizationId edOrgs
            ON ssa.SchoolId = edOrgs.TargetEducationOrganizationId
WHERE edOrgs.SourceEducationOrganizationId = 255901
```

gmcelhanon (commented):
Since I opened the PR, I can't Request Changes, but I am requesting a change here.

gmcelhanon:
@bptillman I approve of the changes, but I opened the original PR and cannot merge it. Please feel free to merge when ready.

bptillman (approved):

PR #448, opened 2022-03-02
[ODS-5255] Confirm that all ODS Platform projects are updated to net6

AxelMarquez:


tests/EdFi.Ods.Api.IntegrationTests/OrganizationDepartmentTests.cs
```diff
@@ -243,16 +243,14 @@ public void When_updating_organization_departments_parent_elsewhere_in_hierarchy
             
             EducationOrganizationHelper.ShouldContainTuples(Connection, expectedTuplesAfterMovedDown);
             
-            afterMovedDownCount.ShouldBe(afterCreationCount + 6,
-                () =>
-                {
-                    return $"The following additional tuples were found: {Environment.NewLine}"
-                    + string.Join(Environment.NewLine, 
-                        EducationOrganizationHelper.GetExistingTuples(Connection)
+            afterMovedDown.Count.ShouldBe(afterCreationCount + 6,
```
AxelMarquez, line 246: Shouldly v4 deprecated `Func<string> customMessage overloads`. We have a roundtrip to the DB to build the error message, however, the existing variable `afterMovedDown` already has the values that we need, so I used it instead.

bptillman (approved):

bptillman (approved):

PR #453, opened 2022-03-09
[ODS-5298] Improve Error Reporting for Bad Connection String

AxelMarquez:
There are 2 main changes in this PR:

1. Fix the database initializers to throw an error if the databases don't exist instead of trying to create them (see comments below).
2. Since the `SecurityRepository` hit the database in the constructor, if the database was invalid the error message was encapsulated in a `DependencyResolutionException` (from Autofac). This made the error a bit cryptic so I added lazy initialization to the repository.

### DIFF
--- a/Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs
+++ b/Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs
@@ -19,7 +19,8 @@ namespace EdFi.Admin.DataAccess.Contexts
         protected UsersContext(string connectionString)
             : base(connectionString)
         {
-            Database.SetInitializer(new ValidateDatabase<UsersContext>());
+            Database.SetInitializer(new ValidateDatabase<SqlServerUsersContext>());
+            Database.SetInitializer(new ValidateDatabase<PostgresUsersContext>());
         }
         public const string UserTableName = "Users";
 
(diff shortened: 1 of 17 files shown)


Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs
```diff
@@ -19,7 +19,8 @@ public abstract class UsersContext : DbContext, IUsersContext
         protected UsersContext(string connectionString)
             : base(connectionString)
         {
-            Database.SetInitializer(new ValidateDatabase<UsersContext>());
+            Database.SetInitializer(new ValidateDatabase<SqlServerUsersContext>());
```
AxelMarquez, line 22: The current version of EF requires the concrete implementation of the Initializer context, note that the initializers weren't being executed because of this, so EF fell back to the default behavior of creating the database if it didn't exist; and it failed to do so because of a cascade delete validation.

This made the error message misleading; the users didn't know that the actual error was an invalid database name in the connection string.

Application/EdFi.Admin.DataAccess/Utils/ValidateDatabase.cs
```diff
@@ -22,12 +22,7 @@ public void InitializeDatabase(TContext context)
             {
                 if (!context.Database.Exists())
                 {
-                    throw new ConfigurationErrorsException("Admin database does not exist.");
-                }
-
-                if (!context.Database.CompatibleWithModel(true))
```
AxelMarquez, line 28: I've removed this validation since we no longer use Code First.

bptillman (approved):

PR #454, opened 2022-03-14
ODS-5303 - Move Admin.DataAccess build to GithubActions

bptillman:
This creates two new GithubActions, one workflow that can be manually triggered and a second that gets triggered on a pull request to main. The manually triggered one mimics what exists in TeamCity right now to build, test, pack, and publish, while the pull request one will only run a build and test.

You will not see the ability to manually run the github workflow since it has to exist in main first before it will show up there, so to test these workflows out I forked ODS, applied the workflows in the main branch and then tested them there in the fork. But you will be able to see the pull request workflow building and passing as part of the checks here.

You will also notice how 10 is being added to the run number for determining the revision in the version for packing and publishing. This is due to the existing TC build having published all the way to to version 5.4.10, so we do not want to override that number.

### DIFF
--- a/.github/workflows/shared.yml
+++ b/.github/workflows/shared.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
     - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
     - name: Setup .NET
-      uses: actions/setup-dotnet@dc30e5defeb4a0047e149c684eab597f5f37f67d # v1
+      uses: actions/setup-dotnet@v1
       with:
         dotnet-version: 6.0.x
     - name: build
--- a/.github/workflows/shared.yml
+++ b/.github/workflows/shared.yml
@@ -1,4 +1,4 @@
-name: Shared workflow
+name: EdFi.Admin.DataAccess Shared workflow
 
 on:
   workflow_call:
--- a/.github/workflows/shared.yml
+++ b/.github/workflows/shared.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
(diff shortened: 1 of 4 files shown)


semalaiappan (changes requested):
Overall looks good .  These 2 workflow names requires to have prefex with 'EdFi.Admin.DataAccess' so that to identify the builds from Github Actions build .  

Note : we need to have task to remove 10 from build counter as per part 5.4 release task as part of post activities

bptillman:
> Overall looks good . These 2 workflow names requires to have prefex with 'EdFi.Admin.DataAccess' so that to identify the builds from Github Actions build .
> 
> Note : we need to have task to remove 10 from build counter as per part 5.4 release task as part of post activities

[ODS-5346](https://tracker.ed-fi.org/browse/ODS-5346) has been added to tracking removing the 10 from the build counter post release

.github/workflows/shared.yml
```diff
@@ -0,0 +1,45 @@
...
+  AZURE_ARTIFACT_URL: "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json"
+  AZURE_ARTIFACT_NUGET_KEY: ${{ secrets.AZURE_ARTIFACTS_PERSONAL_ACCESS_TOKEN }}
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
vimayya, line 23: @bptillman could you please use commit hash instead of version number https://techdocs.ed-fi.org/display/EDFIODS/Guidelines+for+Use+of+GitHub+Actions

semalaiappan (changes requested):
Warning: Unexpected input(s) 'dotnet-version', valid inputs are ['version'] in Setup.NET 
Please use version instead of dotnet-version
Reference https://github.com/actions/setup-dotnet/issues/90

semalaiappan (approved):

PR #459, opened 2022-03-22
[ODS-5305] Ed-Fi Demo Sandbox Refresh is not working

ericstockdale:


Application/EdFi.Ods.Sandbox/Provisioners/SqlServerSandboxProvisioner.cs
```diff
@@ -97,6 +97,17 @@ async Task BackUpAndRestoreSandbox()
 
                         string logicalName = null;
 
+                        _logger.Debug($"dropping {newDatabaseName} it it exists.");
+                        var dropDbIfExists = $@"
```
vimayya, line 101: @ericstockdale Would it be better to move this logic to DeleteSandboxes and call it from ResetSandboxClient before Provisioning sandbox? Could you verify the logic for PostgreSQL as well.

vimayya, line 101: @ericstockdale Would it be better to move this logic to DeleteSandboxes and call it from ResetSandboxClient before Provisioning sandbox? Could you verify the logic for PostgreSQL as well.

ericstockdale: Done

bptillman (approved):

PR #473, opened 2022-03-30
[ODS-5308] Clearing expired tokens in background could potentially allow access via expired tokens

gmcelhanon:
Adds support for returning the expiration time of the token from the GetClientForToken function in SQL Server and PostgreSQL, checking immediately for expired tokens (and when the token is subsequently retrieved from cache) using C# logic.

Also, refines and reduces the amount of code involved in the authentication handler that was written for ASP.NET Core integration.

### DIFF
--- a/Artifacts/MsSql/Structure/Admin/0090-Add-Expiration-to-GetClientForToken.sql
+++ b/Artifacts/MsSql/Structure/Admin/0090-Add-Expiration-to-GetClientForToken.sql
@@ -1,0 +1,55 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+SET QUOTED_IDENTIFIER ON;
+GO
+SET ANSI_NULLS ON;
+GO
+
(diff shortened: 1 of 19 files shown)


Artifacts/MsSql/Structure/Admin/0090-Add-Expiration-to-GetClientForToken.sql
```diff
@@ -0,0 +1,55 @@
...
+        acaeo.[ApiClient_ApiClientId] = cat.[ApiClient_ApiClientId]
+    LEFT OUTER JOIN [dbo].[ApplicationEducationOrganizations] aeo ON
+        aeo.[ApplicationEducationOrganizationId] = acaeo.[ApplicationEducationOrganization_ApplicationEducationOrganizationId]
+		AND (cat.Scope IS NULL OR aeo.EducationOrganizationId = CONVERT(int, cat.Scope))
+    LEFT OUTER JOIN [dbo].[ProfileApplications] ap on
+        ap.[Application_ApplicationId] = app.[ApplicationId]
+    LEFT OUTER JOIN [dbo].[Profiles] p on
+        p.[ProfileId] = ap.[Profile_ProfileId]
+    LEFT OUTER JOIN [dbo].[ApiClientOwnershipTokens] acot ON
+        ac.ApiClientId = acot.ApiClient_ApiClientId
```
semalaiappan, line 54: WHERE cat.Expiration > GETUTCDATE() is missing here

semalaiappan (changes requested):

semalaiappan (approved):

PR #482, opened 2022-04-20
[ODS-5350] - Move LoadTools builds to GithubActions

bptillman:
This is similar to previous additions where EdFi.Admin.DataAccess and EdFi.Security.DataAccess was moved from TeamCity to Github.

Two additional changes made as part of this work were:

- Removing the PackageName script parameter from being required, since the build and test steps of the script do not need them and only the pack and publish did.
- Adding a test category filtering parameter for the test step. The LoadTools build skipped the RunManually tests so a way to do that here was needed as well.

Due to an existing issue, [ODS-5391](https://tracker.ed-fi.org/browse/ODS-5391) the test step of the builds will fail right now. So you can either:
1. Comment out `MappingTests_for_EdFi_models()` in MapperTests OR
2. Make the following changes to the JSON for the test to get it passing, which is the work that would need to be done for ODS-5391:
  electronicMailType > electronicMailTypeDescriptor
  personalInformationVerificationType > personalInformationVerificationDescriptor
  identificationDocumentUseType > identificationDocumentUseDescriptor
  raceType > raceDescriptor

Once this PR is merged then [this PR](https://github.com/Ed-Fi-Alliance/Ed-Fi-TeamCity-Configs/pull/47) should be reviewed for removing the builds from the TeamCity Kotlin configuration.

### DIFF
--- a/.github/workflows/edFi.loadtools manual.yml
+++ b/.github/workflows/edFi.loadtools manual.yml
@@ -1,0 +1,63 @@
+name: EdFi.LoadTools Manually triggered build
+
+on:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+
+env:
+  INFORMATIONAL_VERSION: "5.4"
+  CONFIGURATION: "Release"
+  AZURE_ARTIFACT_URL: "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json"
+  AZURE_ARTIFACT_NUGET_KEY: ${{ secrets.AZURE_ARTIFACTS_PERSONAL_ACCESS_TOKEN }}
+  VSS_NUGET_EXTERNAL_FEED_ENDPOINTS : '{"endpointCredentials": [{"endpoint": "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json","password": "${{ secrets.AZURE_ARTIFACTS_PERSONAL_ACCESS_TOKEN }}"}]}'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 6.0.x
+    - name: build
+      run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/edFi.loadtools manual.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
```
stephenfuqua, line 26: Please change to version hash instead of `@v1`

.github/workflows/edFi.loadtools manual.yml
```diff
@@ -0,0 +1,63 @@
...
+        dotnet-version: 6.0.x
+    - name: build
+      run: |
+        .\build.ps1 build -Configuration ${{ env.CONFIGURATION }} -InformationalVersion ${{ env.INFORMATIONAL_VERSION}} -BuildCounter ${{ github.run_number }} -Solution "Utilities/DataLoading/LoadTools.sln"
+      shell: pwsh
+    - name: test
+      run: |
+        .\build.ps1 test -Configuration ${{ env.CONFIGURATION }} -Solution "Utilities/DataLoading/LoadTools.sln" -TestFilter "RunManually"
+      shell: pwsh
+    - name: pack
```
stephenfuqua, line 37: I see several "pack steps" - would it make sense to be more specific, so that the messages in the GitHub logs have more information?

bptillman: These have all been updated to be given better unique names for each one

.github/workflows/edFi.loadtools manual.yml
bptillman: I forgot what the issue was before with why I did not do this, but now I made the change to see it again. With a hash for this it shows a warning in the build output (but will still complete the run) as you can see here: https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/runs/6159523661?check_suite_focus=true#step:3:1

This same warning does not come up on the checkout action though when using a hash, and I found nothing mentioned in the docs around the Github action for why the warning would be coming up.

Related, I also noticed now that the default workflow template references setup-dotnet@v2, so should I go ahead and move this over to the hash for v2?

stephenfuqua: Perhaps you grabbed the hash value for 1.0.0, whereas @v1 would run the latest @v1.y.z?

Yes, please try out v2.

ericstockdale (approved):

PR #491, opened 2022-04-27
ODS-5349 ODS-5351  GitHub Actions: Setup EdFi.Common build  GitHub Actions: Setup EdFi.Ods.Api build 

semalaiappan:


Application/EdFi.Ods.Api/EdFi.Ods.Api.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
bptillman, line 1: With a few modifications to the build script, adding a solution file for this is not needed to get it this project building and being packaged. In the original teamcity build we call `dotnet build` on `%project.file.sln%` but when you dig further into the parameters it is not actually a solution file and is instead `Ed-Fi-ODS/application/EdFi.Ods.Api/EdFi.Ods.Api.csproj`.

So what could be done for EdFi.Ods.Api is to not pass a value for the Solution parameter and only the projectFile parameter and then at the start of the build script check if Solution is empty, and if so, replace it with projectFile. This way then the calls to dotnet clean, build, and test would still work the way they are now and dotnet pack would remain unchanged since its using the projectFile already.

bptillman, line 1: With a few modifications to the build script, adding a solution file for this is not needed to get it this project building and being packaged. In the original teamcity build we call `dotnet build` on `%project.file.sln%` but when you dig further into the parameters it is not actually a solution file and is instead `Ed-Fi-ODS/application/EdFi.Ods.Api/EdFi.Ods.Api.csproj`.

So what could be done for EdFi.Ods.Api is to not pass a value for the Solution parameter and only the projectFile parameter and then at the start of the build script check if Solution is empty, and if so, replace it with projectFile. This way then the calls to dotnet clean, build, and test would still work the way they are now and dotnet pack would remain unchanged since its using the projectFile already.

.github/workflows/edFi.common manual.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
```
bptillman, line 27: In the last PR for these Github Actions we determined that we can use the hash for setup-dotnet@v2, so please update this and the others in this PR to use that hash, which can be found in `edFi.loadTools manual.yml`

bptillman, line 27: In the last PR for these Github Actions we determined that we can use the hash for setup-dotnet@v2, so please update this and the others to use that hash, which can be found in edFi.loadTools manual.yml

.github/workflows/edFi.common manual.yml
```diff
@@ -0,0 +1,48 @@
...
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+
+env:
+  INFORMATIONAL_VERSION: "5.4"
+  BUILD_INCREMENTER: "10"
```
bptillman, line 13: I believe this one will need to be 11 since there is currently a 5.4.11 out in TeamCity right now, so we'd need this to be 11 so that 11 + first build 1 would be 12.

bptillman, line 13: I believe this one will need to be 11 since there is currently a 5.4.11 out in TeamCity right now, so we'd need this to be 11 so that 11 + first build 1 would be 12.

.github/workflows/edFi.ods.api manual.yml
```diff
@@ -0,0 +1,48 @@
...
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
+
+env:
+  INFORMATIONAL_VERSION: "5.4"
+  BUILD_INCREMENTER: "10"
```
bptillman, line 13: This does not need to be 10 and instead could be 3, since the last build on TeamCity for this is 5.4.3, so we'd want the next in Github to be 5.4.4

bptillman, line 13: This does not need to be 10 and instead could be 3, since the last build on TeamCity for this is 5.4.3, so we'd want the next in Github to be 5.4.4

bptillman (changes requested):

.github/workflows/edFi.common pullrequest.yml
```diff
@@ -0,0 +1,30 @@
+name: EdFi.Common Pull request build and test
+
+on:
+  pull_request:
+    branches: [main]
+
+env:
+  INFORMATIONAL_VERSION: "5.4"
+  BUILD_INCREMENTER: "10"
```
bptillman, line 9: I just realized, but since this is the pull request and no packages get made for this one, the env variable for it here, along with the passing of the value to the build script can be removed for this pull request and the edfi.ods.api pull request workflow.

bptillman (changes requested):

Application/EdFi.Ods.Api/EdFi.Ods.Api.sln
bptillman: Per a conversation we had in standup, this suggestion is no longer being done and the solution file will be kept.

.github/workflows/edFi.common manual.yml
semalaiappan: yes updated . Thanks

.github/workflows/edFi.common manual.yml
semalaiappan: yes updated . Thanks

.github/workflows/edFi.ods.api manual.yml
semalaiappan: yes updated . Thanks

semalaiappan: yes updated . Thanks

Application/EdFi.Ods.Api/EdFi.Ods.Api.sln
semalaiappan: we agreed to use solution file

.github/workflows/edFi.common pullrequest.yml
semalaiappan: yes updated . Thanks

.github/workflows/edFi.common pullrequest.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
```
bptillman, line 19: It appears this one still references v1, and so does `edfi.ods.api pullrequest.yml` so could you please update both of these to reference the v2 hash?

bptillman (changes requested):
This is looking good. Just one last change to make to have all workflows match the v2 hash for setup-dotnet and then this will be good to be approved and merged.

.github/workflows/edFi.common pullrequest.yml
semalaiappan: I fixed it for all builds

bptillman (approved):

PR #517, opened 2022-06-23
[ODS-5462] Single-item authorization of OrganizationDepartment should use ParentEducationOrganizationId

AxelMarquez:


Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
```diff
@@ -102,13 +111,31 @@ protected RelationshipsAuthorizationStrategyBase()
                 Filters = filters,
                 Operator = FilterOperator.Or
             };
-            
+
+            string GetNonRoleNamedProperty(string propertyName)
+            {
+                //Look for the corresponding non - role named property on the model using its DefiningConcreteProperty
+
+                var schema = authorizationContext.Type.GetCustomAttribute<SchemaAttribute>(false)?.Schema;
```
gmcelhanon, line 119: This code is going to be executed _a lot_, so we need to try to make sure we are doing it in a performant way. Reflection is slow, so I think that we should introduce a private `ConcurrentDictionary` and the use of the `GetOrAdd` method here to remember the answers and avoid the reflection hit on every authorization call. The answers are also not specific to each entity, so that logic is valuable for first determining the "originating" name the first time, but no need to incorporate that into the key -- just use the (possibly) role-named property as the key, and the resolved non-role-named property as the value.

Ultimately, we only have about 10 total property name entries to resolve... Student/Staff/Parent USIs, and the EdOrgId, and the EdOrg subtypes, and the occasional role-named EdOrgId-derived values (e.g. ParentEducationOrganizationId, and soon, the ResponsibilitySchoolId in DisciplineIncident).

gmcelhanon (changes requested):
Code enhancement requested... but I still need to review the Postman tests as well.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
```diff
@@ -116,20 +116,26 @@ string GetNonRoleNamedProperty(string propertyName)
...
+
+                    if (!_domainModel.EntityByFullName.TryGetValue(expectedEntityFullName, out var entity))
+                    {
+                        throw new Exception($"Unable to locate entity '{expectedEntityFullName}' in the model.");
+                    }
 
-                return property.DefiningConcreteProperty.PropertyName;
+                    if (!entity.PropertyByName.TryGetValue(propertyName, out var property))
+                    {
+                        throw new Exception($"Property {propertyName} not found in entity '{expectedEntityFullName}'.");
```
gmcelhanon, line 134: Add single quotes around `{propertyName}` in the message.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
```diff
@@ -116,20 +116,26 @@ string GetNonRoleNamedProperty(string propertyName)
             {
                 //Look for the corresponding non - role named property on the model using its DefiningConcreteProperty
 
-                var schema = authorizationContext.Type.GetCustomAttribute<SchemaAttribute>(false)?.Schema;
-                var expectedEntityFullName = new FullName(schema, authorizationContext.Type.Name);
-
-                if (!_domainModel.EntityByFullName.TryGetValue(expectedEntityFullName, out var entity))
+                return _nonRoleNamedByRoleNamedPropertyName.GetOrAdd(propertyName, _ =>
```
gmcelhanon, line 119: Consider using the parameter to the delegate rather than the outer `propertyName` variable to avoid unnecessary use of a closure.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
```diff
@@ -141,6 +147,7 @@ string GetAuthorizationFilterName(string subjectEndpointName, string authorizati
         }
 
         private readonly ConcurrentDictionary<(string, string), string> _filterNameBySubjectAndPathModifier = new();
+        private readonly ConcurrentDictionary<string, string> _nonRoleNamedByRoleNamedPropertyName = new();
```
gmcelhanon, line 150: Let's rename this to `_nonRoleNamedByRawPropertyName` because otherwise it reads like the key is a role-named property, but really it's more of a _possibly_ role named property. I think "raw" property name is a better term here.

gmcelhanon (changes requested):

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
AxelMarquez: Added a cache around it

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
AxelMarquez: Fixed it

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
AxelMarquez: Renamed

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/RelationshipsAuthorizationStrategyBase.cs
AxelMarquez: Fixed it

gmcelhanon (approved):

PR #521, opened 2022-06-28
[ODS-5481] Improve use of X-Forwarded-XYZ headers (main)

stephenfuqua:


Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
```diff
@@ -59,7 +59,8 @@ public abstract class DataManagementControllerBase<TResourceReadModel, TResource
         private readonly IRESTErrorProvider _restErrorProvider;
         private readonly int _defaultPageLimitSize;
         private readonly bool _useProxyHeaders;
-
+        private readonly string _defaultForwardingHostServer;
```
AxelMarquez, line 62: Why were these 2 variables introduced? seems that they aren't used

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
```diff
@@ -43,45 +43,56 @@ public static string Scheme(this HttpRequest request, bool useProxyHeaders = fal
                 scheme = proxyHeaderValue;
             }
 
-            return scheme;
+            // The x-forwarded-proto header can contain a single originating protocol or, in some 
+            // cases, multiple protocols. Seee ODS-5481 for more information. We care about the
```
AxelMarquez, line 47: Typo in ``Seee``

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
```diff
@@ -43,45 +43,56 @@ public static string Scheme(this HttpRequest request, bool useProxyHeaders = fal
...
+                return request.Host.Port ?? getDefaultPort();
             }
 
+            // Try to extract a X-Forwarded-Port value
             request.TryGetRequestHeader(HeaderConstants.XForwardedPort, out string proxyHeaderValue);
 
-            return !int.TryParse(proxyHeaderValue, out int port)
-                ? request.Host.Port ?? defaultPortForScheme
-                : port;
+            if (proxyHeaderValue != null)
```
AxelMarquez, line 84: Should we check for ``IsNullOrWhiteSpace`` here as well? To handle the scenario where the header is defined with an empty value.

Application/EdFi.Ods.Api/Controllers/VersionController.cs
```diff
@@ -126,7 +129,7 @@ public IActionResult Get()
 
                 if (_apiSettings.IsFeatureEnabled(ApiFeature.XsdMetadata.GetConfigKeyName()))
                 {
-                    urlsByName["xsdMetadata"] = Request.RootUrl(useReverseProxyHeaders) + "/metadata/" +
+                    urlsByName["xsdMetadata"] = rootUrl + "/metadata/" +
```
AxelMarquez, line 132: There are 3 places that need to be updated to reference ``rootUrl`` as well, see ``changeQueries``, ``composites`` and ``identity`` sections a few lines below.

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
```diff
@@ -7,20 +7,20 @@
...
-using EdFi.Ods.Common.Extensions;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Primitives;
 
 namespace EdFi.Ods.Api.Extensions
 {
     public static class HttpRequestExtensions
     {
-        public static string RootUrl(this HttpRequest request, bool useProxyHeaders = false)
+        public static string RootUrl(this HttpRequest request, bool useProxyHeaders = false,
```
AxelMarquez, line 17: There are a couple of other calling sites that have to specify these new parameters, for example, `OpenApiMetadataController`; since it doesn't pass the parameters, it falls back to the default values and generates incorrect endpoint urls in the metadata (used by Swagger).

I recommend making the parameters mandatory so the compiler shows all the places that need to be updated.

AxelMarquez (changes requested):

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
stephenfuqua: Apparently it was a dead end. I should be able to remove it - either that, or I forgot to use these in an obscure code path.

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
stephenfuqua: Possibly. Will look into it.

stephenfuqua: This does behave correctly in that scenario. I've added Postman tests and covered this in unit tests as well.

Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
stephenfuqua: Removed

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
stephenfuqua: done

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
stephenfuqua: done

Application/EdFi.Ods.Api/Controllers/VersionController.cs
stephenfuqua: done, thanks for pointing those out - i missed them in the cherry pick, as those lines don't exist in 5.3.

Application/EdFi.Ods.Api/Extensions/HttpRequestExtensions.cs
stephenfuqua: I've addressed the other classes that call `RootUrl`.

Application/EdFi.Ods.Common/Configuration/ApiSettings.cs
```diff
@@ -50,6 +53,27 @@ public ApiSettings()
...
+            {
+                return _defaultForwardingHostServer ?? "localhost";
+            }
+            set
+            {
+                _defaultForwardingHostServer = value;
+            }
+        }
+
+        public int DefaultForwardingHostPort
```
AxelMarquez, line 68: There is an error if the api is started with `DefaultForwardingHostPort = ""` (and `UseReverseProxyHeaders = false`), since this property is an _int_ it always expects a valid number during startup.

I recommend changing it to a nullable _int_ (`int?`) and instead of relying on fallback values (like _localhost_ and port _80_ here, or port _443_ on _HttpRequestExtensions_ methods parameters) let them default to null, and if ``UseReverseProxyHeaders = true`` and any of the related _DefaultForwardingHostPort_ or _DefaultForwardingHostServer_ is invalid throw an error instructing the user to check his appsettings.

Thoughts?

AxelMarquez (changes requested):

Application/EdFi.Ods.Common/Configuration/ReverseProxySettings.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public ReverseProxySettings(bool? useReverseProxyHeaders, string defaultForwardingHostServer, int? defaultForwardingHostPort)
+        {
+            if (useReverseProxyHeaders.HasValue
+                && useReverseProxyHeaders.Value
+                && (string.IsNullOrWhiteSpace(defaultForwardingHostServer)
+                    || !defaultForwardingHostPort.HasValue)
+                )
+            {
+                throw new ArgumentException(ForwardingConfigurationError);
```
stephenfuqua, line 23: Here's the runtime error if you don't have the right default settings

AxelMarquez (approved):

PR #540, opened 2022-07-24
[ODS-5512] Improve matching between dependencies endpoint and SDK resources

AxelMarquez:


Utilities/DataLoading/EdFi.LoadTools/SmokeTest/SdkTests/DestructiveMethodsGenerator.cs
```diff
@@ -29,7 +29,9 @@ public class DestructiveMethodsGenerator : ITestGenerator
             _dependenciesRetriever = dependenciesRetriever;
             _dependenciesSorter = dependenciesSorter;
         }
-        
+
+        protected ILog Log => LogManager.GetLogger(GetType().Name);
```
bptillman, line 33: I was originally going to ask why you went with declaring the logger as `protected` since we typically have it `private readonly`, but then I found a few instances of it being protected, so this can stay.

bptillman (approved):
Rebased this on latest from main and updated the three relevant packages. After that, changes look good to merge.

PR #550, opened 2022-08-08
[ODS-5525] YearSpecific API GETs are broken

AxelMarquez:


semalaiappan (approved):
Deployed in Staging for  Year-Specific & SharedInstance .  API is working fine in both urls .

PR #551, opened 2022-08-12
[ODS-5477] - Setup workflow to trigger implementation repo workflows on PRs

bptillman:
A new workflow has been setup utilizing the repository dispatch event, so that any PR activity on the ODS repo will trigger the new InitDev workflows created in the Implementation repo, ensuring that commits to both repos trigger the same workflows.

### DIFF
--- a/.github/workflows/InitDev Implementation.yml
+++ b/.github/workflows/InitDev Implementation.yml
@@ -1,0 +1,22 @@
+name: Trigger InitDev workflows in Implementation repo
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


AxelMarquez:
Open Api Generator got updated to v6.0.1 in ODS-5503. In this PR I've also fixed the breaking change that this new version introduced, it added new optional parameters to the API methods, so I now pass ``Type.Missing`` while doing the reflective call to ignore these new parameters.

semalaiappan (approved):

PR #577, opened 2022-10-06
[ODS-5601] Issues to do with Profiles on TEA project

AxelMarquez:


AxelMarquez:
Closed since the change is going to be included in https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/580

PR #580, opened 2022-10-07
[ODS-5601] Issues to do with Profiles on TEA project (only some references with unified key included)

gmcelhanon:
Fixed code generation issue where when a reference which has unified key properties is included, while the other reference(s) are not included, compilation would fail due to the missing private "implicit" reference artifacts.

### DIFF
--- a/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/ResourceProfileData.cs
+++ b/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/ResourceProfileData.cs
@@ -172,6 +172,13 @@ namespace EdFi.Ods.CodeGen.Generators.Resources
 
         public bool IsIncluded(ResourceClassBase resource, Reference reference)
         {
+            return IsIncluded(resource, reference, out var ignored);
+        }
+        
+        public bool IsIncluded(ResourceClassBase resource, Reference reference, out bool implicitOnly)
+        {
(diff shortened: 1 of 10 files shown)


Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/ResourceProfileData.cs
```diff
@@ -200,20 +193,41 @@ public bool IsIncluded(ResourceClassBase resource, Reference reference, out bool
...
-                    .Except(new[] { reference })
-                    .SelectMany(r => r.Properties);
+                bool isSubsetOfAnotherIncludedReference = 
+                    // All references (except the current reference)
+                    resource.References.Except(new[] { reference })
+
+                    // Inspect other included references
+                    .Where(r => IsIncluded(resource, r))
+
+                    // Exclude the current reference
```
AxelMarquez, line 221: Seems that this comment is no longer needed, it also appears a few lines above.

AxelMarquez (approved):

Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/ResourceProfileData.cs
AxelMarquez: Will get fixed in a different PR.

PR #582, opened 2022-10-11
ODS-5582 Remove Ownership Based Test claim set from EdFi_Secuirty

semalaiappan:


Artifacts/MsSql/Data/Security/2090-RemoveOwnershipBasedTestClaimSet.sql
```diff
@@ -0,0 +1,31 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+
+IF DB_NAME() = 'EdFi_Security'
```
bptillman, line 7: So this works in the default scenario, as long as the database is named EdFi_Security.

But if it was deployed under a different name, for example EdFi_Ods_Security, by modifying the connection string in appsettings.json this would not be run against it. Looking at our DBUp configuration [here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-Databases/blob/main/src/EdFi.Db.Deploy/UpgradeEngineFactories/SqlServerUpgradeEngineFactory.cs#L24) in DbDeploy, I can see that we do not allow variable substitution which would be one way we could try to work around this by swapping out a variable in the script file for the actual name of the database we parse from the connection string. If we were to allow variable substitution that could work, but that would be more work than what was initially asked of this ticket.

Thoughts @semalaiappan @vimayya ?

bptillman (approved):

PR #592, opened 2022-10-20
ODS-5562 Move EdFi.Ods.Minimal.Template.TPDM.PostgreSQL Package build to Github Actions

semalaiappan:
https://intedfitools1.msdf.org/buildConfiguration/EdFi_SharedResources_OdsPlatformChangeQuery_OdsApi_Packages_EdFiOdsMinimalTemplateTpdmPostgreSQL?branch=&buildTypeTab=overview&mode=builds#all-projects  you can refer this build to verfiy each step from teamcity

### DIFF
--- a/.github/workflows/Packages - EdFi.Ods.Minimal.Template.TPDM.PostgreSQL.yml
+++ b/.github/workflows/Packages - EdFi.Ods.Minimal.Template.TPDM.PostgreSQL.yml
@@ -54,6 +54,11 @@ jobs:
           repository: Ed-Fi-Alliance-OSS/Ed-Fi-Standard
           path: Ed-Fi-Standard/
           ref: ${{ env.EDFI_STANDARD_REFERENCE }}
+    - name: Checkout Ed-Fi-TPDM-Artifacts
+      uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2
+      with:
+          repository: Ed-Fi-Alliance-OSS/Ed-Fi-TPDM-Artifacts
+          path: Ed-Fi-TPDM-Artifacts/          
     - name: Is pull request branch exists in Ed-Fi-ODS-Implementation
       working-directory: ./Ed-Fi-ODS/
       shell: powershell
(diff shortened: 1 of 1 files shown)


.github/workflows/Packages - EdFi.Ods.Minimal.Template.TPDM.PostgreSQL.yml
```diff
@@ -0,0 +1,110 @@
...
+      shell: powershell
+    - name: Install-credential-handler
+      run: iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) } -AddNetfx"
+      shell: powershell
+    - name: publish 
+      working-directory: ./Ed-Fi-ODS/
+      run: |
+        .\build.githubactions.ps1 publish -InformationalVersion ${{ env.INFORMATIONAL_VERSION }} -BuildCounter ${{ github.run_number }} -BuildIncrementer ${{env.BUILD_INCREMENTER}} -NuGetApiKey ${{ env.AZURE_ARTIFACT_NUGET_KEY }} -EdFiNuGetFeed ${{env.AZURE_ARTIFACT_URL}} -PackageName  "EdFi.Suite3.Ods.Minimal.Template.TPDM.Core.PostgreSQL"
+      shell: powershell
+    - name: Upload EdFi.Ods.Minimal.Template.TPDM.PostgreSQL Artifacts
```
bptillman, line 105: The previous build in TeamCity also had a results.zip file that contained the logs from the Bulk Load Client and from the Test Harness, whether it was a successful or failed build. Should those be added as artifacts of this build as well? This looks to be something we missed with porting over the other template builds as well so if we want to add it, we should create a ticket to add it back to the other template builds here in Github as well.

image

bptillman (commented):
Just one comment I had, related to keeping the build artifacts the same as what we had in TeamCity. Let me know your thoughts.

.github/workflows/Packages - EdFi.Ods.Minimal.Template.TPDM.PostgreSQL.yml
semalaiappan: image

Just having it in  NugetPackages.Artifacts zip is enough I think but definitely need logfile file and also testharnesslog file .

bptillman (approved):

PR #594, opened 2022-10-21
ODS-5587 Integrate latest 4.0 Data Standard Work

semalaiappan:


.github/workflows/Packages - EdFi.Ods.Minimal.Template.PostgreSQL.yml
```diff
@@ -13,7 +13,7 @@ on:
   repository_dispatch:
     types: [triggered-from-implementation-repo]
 env:
-  EDFI_STANDARD_REFERENCE: "4.0.0-a"
```
bptillman, line 16: This comment applies to this workflow and all the other workflows with this change:

- Could we instead keep this env var and just change the value to `main`? This way the next time when we need to point this to a tag, we just update it to the appropriate value instead of having it add it back in two places
- Additionally, do we need the Is pull request step? I thought in TeamCity we usually kept the DS repo pointed to a tag or main.

Application/EdFi.Ods.Standard/Artifacts/Metadata/ApiModel.json
```diff
@@ -63399,7 +63399,7 @@
         "schema": "edfi",
         "name": "FK_DisciplineActionStudentDisciplineIncidentAssociation_DisciplineAction"
       },
-      "cardinality": "OneToOneOrMore",
+      "cardinality": "OneToZeroOrMore",
```
bptillman, line 63402: When I copied in ApiModel.json from Metadata it looks to have a few more changes compared to this one here. For example, I see `EducationOrganizationAssociationTypeDescriptor` being added in mine locally, which was in the sql file changes as well. But it's not showing that change in the PR here. Another example is `StudentAssessmentEducationOrganizationAssociation` that I do not see here but I believe should be here.
image

bptillman: `ReportedSchoolIdentifier` also shows up as a change, along with a few more FK changes that I saw in SQL but not seen here in Apimodel.json. Could you re-run this and check to see if you see any of these additional changes?

bptillman (changes requested):

.github/workflows/Packages - EdFi.Ods.Minimal.Template.PostgreSQL.yml
semalaiappan: Fixed it .   There is PR in DS repo https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-Standard/pull/142  which excluding fianance.xml since we have bug in that file . Don't Merge it .  https://tracker.ed-fi.org/browse/ODS-5629 needs to be merged first and then next set of DataStandard ODS ticket , we will generate template builds along with finance sample data in ODS

Application/EdFi.Ods.Standard/Artifacts/Metadata/ApiModel.json
semalaiappan: yes  I did . Updated it

bptillman (approved):

PR #595, opened 2022-10-21
ODS-5610 Integration test harness appsettings.json file update to enable Profiles feature

semalaiappan:


gmcelhanon:
All of these feature checks have been removed so that the tests will always run.

PR #608, opened 2022-11-22
ODS-5589 Cleanup Github Actions builds

semalaiappan:


.github/workflows/CodeQL Security Scan.yml
```diff
@@ -48,31 +50,19 @@ jobs:
             path: Ed-Fi-ODS-Implementation/
       - name: Is pull request branch exists in Ed-Fi-ODS-Implementation
         working-directory: ./Ed-Fi-ODS-Implementation/
-        shell: pwsh
+        shell: powershell
```
bptillman, line 53: Similar question here as to the other PRs, but if things worked in pwsh why not leave it as that?

bptillman (changes requested):

.github/workflows/CodeQL Security Scan.yml
semalaiappan: fixed it

PR #611, opened 2022-11-27
[ODS-5654] Composites Open Api metadata doesn't include some properties from the base resource

AxelMarquez:
Composites mutate ([here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/66cd7af54d99145b21e8ae2d9ddaeb37b38bbc6a/Application/EdFi.Ods.Features/Composites/CompositeResourceModelBuilder.cs#L112)) the Resource _Properties_ after the Resource is instantiated.
__allPropertiesRaw_ was being initialized during instantiation, meaning it was missing the changes made by Composites. Wrapping it in a Lazy resolves the issue.

### DIFF
--- a/Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataDefinitionsFactory.cs
+++ b/Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataDefinitionsFactory.cs
@@ -60,35 +60,31 @@ namespace EdFi.Ods.Features.OpenApiMetadata.Factories
                 .Where(x => _openApiMetadataFactoryResourceFilterStrategy.ShouldInclude(x.Resource))
                 .SelectMany(
                     r => new[]
-                    {
-                        new
                         {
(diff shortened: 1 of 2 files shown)


Application/EdFi.Ods.Features/OpenApiMetadata/Factories/OpenApiMetadataDefinitionsFactory.cs
```diff
@@ -60,35 +60,31 @@ private static Schema EtagSchema
...
-                            : null
-                    })
-                .Where(d => d != null)
+                            (
+                                key: _openApiMetadataDefinitionsFactoryNamingStrategy.GetResourceName(r.Resource, r),
+                                schema: CreateResourceSchema(r)
+                            )
+                        }
+                        .Concat(
+                            isChangeQueriesEnabled && !r.IsCompositeResource
```
AxelMarquez, line 70: This is just a slight improvement, don't include Change Queries schemas for Composite since they are not needed. Before this change, the definitions were included but never used.

bptillman (approved):

PR #612, opened 2022-11-30
[ODS-5662] DestructiveSdk smoke tests fail if executed from Linux

AxelMarquez:
The smoke tests use fuzzy search to know how to map properties from the EvaluationRating into the needed EvaluationRatingReference in EvaluationObjectiveRatings.

The fuzzy search constructs candidates separating them with '\\', for example, to resolve EvaluationRatingReference\PersonId, a candidate is `EvaluationRating\PerformanceEvaluationRatingReference\PersonId`, another candidate is `EvaluationRating\Reviewers\PersonId`, and a score is given to each candidate based on how similar the paths are. The candidate with the highest score gets used.

In Linux, the directory separator char is '/' which is treated differently by the [method ](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/66d976f8922ac0f75324a1cdc7c04d06b2e44143/Utilities/DataLoading/EdFi.LoadTools/Engine/ExtensionMethods.cs#L112)that calculates the scores, so different candidates are selected.

With this change, I always use '\\' as property path separator, which is safe since we are not working with directories.

### DIFF
--- a/Utilities/DataLoading/EdFi.LoadTools/Mapping/MapperConfigurationExpression.cs
+++ b/Utilities/DataLoading/EdFi.LoadTools/Mapping/MapperConfigurationExpression.cs
@@ -3,6 +3,7 @@
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
 
+using EdFi.LoadTools.Engine;
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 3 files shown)


Utilities/DataLoading/EdFi.LoadTools/Mapping/MapperConfigurationExpression.cs
```diff
@@ -115,19 +116,21 @@ private static IEnumerable<PropertyInfoEx> GetRecursiveProperties(Type type, str
                     continue;
                 }
 
+                var propertyPath = $"{path}{Constants.PropertyPathSeparator}{propertyInfo.Name}".Trim(Constants.PropertyPathSeparator);
```
AxelMarquez, line 119: Trim() is needed because path can be an empty string and we don't want leading slashes.

bptillman (approved):

PR #616, opened 2022-12-08
ODS-5574 - Add validation for new min and max value properties for integer and decimals

bptillman:
With two newly added fields of minValue and maxValue, the existing logic that creates the RangeAttribute on properties now needed to be updated to return the more restricted range that could be calculated between minValue, maxValue, and the values created from a precision and scale specified.

What that means is that if you had the following values defined:
precision:4
scale: 2
minValue: 1
maxValue: 1000

The resulting more restrictive portion of that we would want to use for the RangeAttribute would end up being from 1 to 99.99, as shown here:
image

Additionally, logic was added for handling this for integers as well when they have min and max value specified.

Since the ApiModel.json has not been updated yet to account for these changes, to test this here is what I did and you can do something similar to test this:
1. Open ApiModel.json and updated the following on the Course entity: MaximumAvailableCredits minValue 1, maxValue 10; MinimumAvailableCredits minValue 1000000, maxValue 10; MinimumAvailableCreditConversion minValue 1, maxValue 10000000
2. Run initdev
3. Execute CodeGen tests and then you should see a diff that shows the new ranges specified should be as follows:
  a. MaximumAvailableCredits Range 1 to 10 (to test picking both min and max value to be the range since its the most restrictive)
  b. MinimumAvailableCredits Range -999999.99 to 10 (to test picking only maxValue since that's more restrictive than the generated max from precision/scale)
  c. MinimumAvailableCreditsConverstion Range 1 to 9999999.99 (to test picking only minValue since that's more restrictive than the generated min from precision/scale)

image


Application/EdFi.Ods.Common/Models/Domain/PropertyType.cs
```diff
@@ -12,7 +13,7 @@ namespace EdFi.Ods.Common.Models.Domain
 {
     public class PropertyType
     {
-        public PropertyType(DbType dbType, int maxLength = 0, int precision = 0, int scale = 0, bool isNullable = false)
+        public PropertyType(DbType dbType, int maxLength = 0, int precision = 0, int scale = 0, bool isNullable = false, decimal? minValue = null, decimal? maxValue = null)
```
gmcelhanon, line 16: No change required here, but just adding a note for posterity.

If we ever need to support min/max values for anything other than integers or decimal values (e.g. dates), we'll probably want to move the min/max elements out of the PropertyType structure (into a new PropertyConstraints structure, along with the isNullable value?) which has now grown beyond its originally intended scope.

gmcelhanon (approved):

PR #635, opened 2023-01-05
[ODS-5675] Include basic Request-Response details in the logs

simpat-jesus:
Tested using Invoke-PostmanIntegrationTests
image


semalaiappan (approved):
I tested with both Error and Info level and both are working fine

PR #636, opened 2023-01-10
[ODS-5664] Relationship-based authorization of PUT/POST/DELETE requests returns 500 status for API client with no ed orgs

gmcelhanon:
Fix SQL used for authorization so that it returns no results rather than cause runtime error due to SQL syntax error.


semalaiappan (changes requested):
Invoke-RebuildSolution step is failing in teamcity & Github Actions builds

semalaiappan (approved):

PR #652, opened 2023-02-08
[ODS-5727] Simplifications to API client experience using Profiles 

semalaiappan:


Application/EdFi.Ods.Api/Filters/EnforceAssignedProfileUsageFilter.cs
```diff
@@ -97,7 +97,7 @@ public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionE
                         ? contentTypes.Readable
                         : contentTypes.Writable)
                     != null)
-            .ToArray();
+            .OrderBy(a=>a).ToArray();
```
gmcelhanon, line 100: Why does this need to be sorted?

Postman Test Suite/Ed-Fi ODS-API Profile Test Suite.postman_collection.json
```diff
@@ -1,9 +1,8 @@
 {
 	"info": {
-		"_postman_id": "c0b4baa3-57b0-441a-856c-b53397e5292d",
+		"_postman_id": "7cb234f7-8ee4-4d00-b45b-35a6cf653730",
```
gmcelhanon, line 3: As a side note, I generally try to reset these changed lines when staging the commit to reduce the noise of unnecessary churn.

Application/EdFi.Ods.Api/Filters/EnforceAssignedProfileUsageFilter.cs
semalaiappan: So that string errorMessage will be same for both database

semalaiappan: string errorMessage = relevantContentTypeUsage == ContentTypeUsage.Readable
                ? $"Based on profile assignments, one of the following profile-specific content types is required when requesting this resource: '{string.Join("', '", assignedProfilesForRequest.Select(p => ProfilesContentTypeHelper.CreateContentType(resourceFullName.Name, p, relevantContentTypeUsage)))}'"
                : $"Based on profile assignments, one of the following profile-specific content types is required when updating this resource: '{string.Join("', '", assignedProfilesForRequest.Select(p => ProfilesContentTypeHelper.CreateContentType(resourceFullName.Name, p, relevantContentTypeUsage)))}'";

semalaiappan: otherwise postgres returns values in different order for assignedProfilesForRequest and throws error in postman test , I need to sorted the assignedProfilesForRequest to get same ordered values to have same error message in response

Postman Test Suite/Ed-Fi ODS-API Profile Test Suite.postman_collection.json
semalaiappan: OK sure . Thanks .  I will do next time .

Application/EdFi.Ods.Api/Filters/EnforceAssignedProfileUsageFilter.cs
gmcelhanon: So then why not just perform the `Sort` down in the `WriteForbiddenResponse` method in the case where you actually need it, instead of adding to the 100% execution path? Move it down to that method.

semalaiappan: I updated it . Thanks

PR #662, opened 2023-02-24
[ODS-5738] Introduce general-purpose caching interceptor and apply to Profiles metadata provider

gmcelhanon:


gmcelhanon:
* Added an overload to the `ExpiringConcurrentDictionaryCacheProvider` that takes an `Action` that is invoked when the cache period expires. This is injected by the registration code to send the `ProfileMetadataCacheExpired` notification out through Mediatr to all handlers.
* Added a couple extension methods to make the syntax around the registration of "ResolvedParameters" in Autofac simpler.
* Implemented the `INotificationHandler<T>` interface on classes (see the `Handle` methods) that do work with the Profile metadata and the store the results in memory. This allows that work to be reset.
* Changed Autofac registrations of the components to register the classes using all their interfaces so that the handlers are registered and Mediatr can find them.
* Added debug logging entries around the resetting and in some cases, the initialization, of Profiles-related work.

Application/EdFi.Ods.Features/OpenApiMetadata/Providers/OpenApiMetadataCacheProvider.cs
```diff
@@ -184,15 +198,26 @@ void AddToCache(IEnumerable<OpenApiContent> openApiContents)
...
                     // we want to force an update if the document has changed.
-                    _openApiMetadataMetadataCache
-                        .AddOrUpdate(
-                            $"{openApiContent.Section}-{openApiContent.Name}",
-                            openApiContent,
-                            (key, existing) => existing.GetHashCode() != openApiContent.GetHashCode()
-                                ? openApiContent
-                                : existing);
+                    newCache.AddOrUpdate(
+                        $"{openApiContent.Section}-{openApiContent.Name}",
```
mjaksn, line 202: The change to the dictionary on 76 above, which results in it being case-sensitive, is causing cache misses when requesting some OpenApi endpoints for some features, including profiles and change queries. The cache misses result in 404 errors being returned when they are requested. An example endpoint where I see this is /profiles/sample-profile-resource-includeonly/swagger.json. 

If the cache dictionary needs to remain as it is, then changing this to be `"{openApiContent.Section}-{openApiContent.Name.ToLowerInvariant()}",` may be the most straightforward solution.

Application/EdFi.Ods.Common/Caching/CacheInterceptor.cs
```diff
@@ -0,0 +1,58 @@
+﻿// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System.Reflection;
+
+namespace EdFi.Ods.Common.Caching;
```
mjaksn, line 8: Can you consolidate the using statements

mjaksn (changes requested):

Application/EdFi.Ods.Common/Caching/CacheInterceptor.cs
gmcelhanon: Done.

Application/EdFi.Ods.Features/OpenApiMetadata/Providers/OpenApiMetadataCacheProvider.cs
gmcelhanon: Changed line 120 from this:

```csharp
var newCache = new ConcurrentDictionary<string, OpenApiContent>();
```
... to this...
```csharp
var newCache = new ConcurrentDictionary<string, OpenApiContent>(StringComparer.OrdinalIgnoreCase);
```

gmcelhanon:
@mjaksn @vimayya Closing this PR because the functionality in this branch consists of [ODS-5738](https://tracker.ed-fi.org/browse/ODS-5738) + [ODS-5757](https://tracker.ed-fi.org/browse/ODS-5757) (a new ticket introducing Mediatr and the support for resetting all work done based on the Profile metadata when the cache period expires).

We will first review and merge the [PR for ODS-5738](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/664), and then rebase the latter commits on this branch as ODS-5757 on the end of `main` branch after the merge.

PR #665, opened 2023-02-27
[ODS-5729] Consider updating API error messages to be more meaningful to the API client 

mjaksn:
This updates the API error message referenced in the ticket to follow the proposed format. I searched for other error messages using the same form as the original message but found none.

### DIFF
--- a/Application/EdFi.Ods.Api/Security/Authorization/AuthorizationBasisMetadataSelector.cs
+++ b/Application/EdFi.Ods.Api/Security/Authorization/AuthorizationBasisMetadataSelector.cs
@@ -289,10 +289,9 @@ public class AuthorizationBasisMetadataSelector : IAuthorizationBasisMetadataSel
                     principal.Claims.SingleOrDefault(c => c.Type == EdFiOdsApiClaimTypes.ClaimSetName)?.Value;
 
                 response.SecurityExceptionMessage =
-                    $@"Access to the resource could not be authorized. Are you missing a claim? This resource can be authorized by the following claims:
-    {string.Join(Environment.NewLine + "    ", authorizingClaimNames)}
-The API client has been assigned the '{apiClientClaimSetName}' claim set with the following resource claims:
-    {string.Join(Environment.NewLine + "    ", apiClientResourceClaims)}";
+                    $@"Access to the resource could not be authorized. The API client has been assigned the '
+                        {apiClientClaimSetName}' claim set. Assign different claim set which includes one of the following claims to access this resource:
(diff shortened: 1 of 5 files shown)


mjaksn:
Reopening this PR

mjaksn:
These changes update the API error messages referenced in the ticket and another with the same form to be more concise and removes non-applicable wording. In addition, tests that include the affected error messages were also updated to align with the new language.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationBasisMetadataSelector.cs
```diff
@@ -289,10 +289,9 @@ void ValidateAuthorizationContext(EdFiAuthorizationContext authorizationContext)
                     principal.Claims.SingleOrDefault(c => c.Type == EdFiOdsApiClaimTypes.ClaimSetName)?.Value;
 
                 response.SecurityExceptionMessage =
-                    $@"Access to the resource could not be authorized. Are you missing a claim? This resource can be authorized by the following claims:
-    {string.Join(Environment.NewLine + "    ", authorizingClaimNames)}
-The API client has been assigned the '{apiClientClaimSetName}' claim set with the following resource claims:
-    {string.Join(Environment.NewLine + "    ", apiClientResourceClaims)}";
+                    $@"Access to the resource could not be authorized. The API client has been assigned the '
```
simpat-jesus, line 292: Current formatting is
```
Access to the resource could not be authorized. The API client has been assigned the '
            ClientClaimSet' claim set. Assign different claim set which includes one of the following claims to access this resource:
            claim one
    claim two
```

fix it to be like
```
Access to the resource could not be authorized. The API client has been assigned the 'ClientClaimSet' claim set.
Assign different claim set which includes one of the following claims to access this resource:
    claim one
    claim two
```

simpat-jesus (changes requested):

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationBasisMetadataSelector.cs
simpat-jesus: Even if the result is serialized to json, we are not sure how clients are parsing the message; so I think is better to return the message as readable as possible, even if that includes new line and carriage return characters once is serialized

mjaksn: OK, I left them in and pushed the change.

simpat-jesus (approved):

PR #699, opened 2023-04-13
[ODS-5796] Modify builds to support multiple standard versions

simpat-jesus:
Builds are passing now

https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/actions/runs/4683667133

https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/actions/runs/4683868293


semalaiappan (changes requested):
Can you please shorten build name like this 
image
Right now Extension Version is not visible 
image
I think Std 4.0.0 Ext 1.0.0 is better naming for build name .  Otherwise every thing looks good

semalaiappan (approved):

PR #704, opened 2023-04-19
[ODS-5752] Authorization strategy for reference data

gmcelhanon:


mjaksn (approved):
I manually verified that applying claims that use the RelationshipsWithEdOrgsAndPeopleInverted authorization strategy will grant clients read access up the hierarchy for course, program, and chartOfAccount resources on mssql and pgsql deployments

PR #747, opened 2023-06-09
[SF-2199] Adding EnableSsl to RedisCacheSettings

nsd-sergio:
https://dellfoundation.lightning.force.com/lightning/r/Case/500Hp00001WdBJVIA3/view

### DIFF
--- a/Application/EdFi.Ods.Common/Configuration/CacheSettings.cs
+++ b/Application/EdFi.Ods.Common/Configuration/CacheSettings.cs
@@ -52,6 +52,7 @@ namespace EdFi.Ods.Common.Configuration
         {
             public string Configuration { get; set; }
             public string Password { get; set; }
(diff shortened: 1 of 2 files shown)


ea-mtenhoor (commented):
This is probably functional, but still doesn't use the built-in parser provided by the Redis module.  There are other connection string options that are still excluded by trying to manually build the configurationOptions object instead of letting the built-in parser do it.

nsd-sergio:
Hi @ea-mtenhoor Technical team is already aware of the 2 possible solutions(one from your side and mine), and they will determine what the best solution is; it could be a combination of both or a completely new one.

Thanks for the feedback

stephenfuqua:
The team had decided to strictly use the parser with a single input string. Even the password option has been removed. Therefore closing the PR. Nonetheless, thank you @nsd-sergio for submitting this

PR #749, opened 2023-06-14
[ODS-5903] Fix the bit value used for the ReadChanges action

gmcelhanon:
Fix the bit value used for the ReadChanges action.


vimayya (approved):

vimayya:
Failing build is a stale build that is no longer applicable and has been removed now. Merging the change.

PR #753, opened 2023-06-19
[ODS-5895] Update Smoke Test to handle Min/Max values and MinLength/MaxLength constraints

vimayya:


Utilities/DataLoading/EdFi.LoadTools/SmokeTest/PropertyBuilders/BaseBuilder.cs
```diff
@@ -82,5 +89,27 @@ protected bool IsRequired(PropertyInfo propertyInfo)
 
             return parameter.required.HasValue && parameter.required.Value;
         }
+
+        protected int BuildRandomNumber(PropertyInfo propertyInfo)
+        {
+            var parameter = _metadataLookup.GetMetadata(propertyInfo);
+            
+            if (parameter.minimum.HasValue && parameter.maximum.HasValue)
```
simpat-jesus, line 97: For GraduationPlan.TotalRequiredCredits, we have a range validation from 0 to 999999.999 but swagger metadata only exposes a minValue of 0, so this will handle required properties with/without both min/max values or only one of the two values provided

semalaiappan (approved):

PR #766, opened 2023-07-03
[ODS-5865] Validate resource properties with meaningful default values

simpat-adam:
This PR introduces a mechanism to validate resource properties whose default value has domain meaning, and therefore must be explicitly assigned as opposed to just accepting the default value when the value is omitted in the JSON payload and model binding occurs.

### DIFF
--- a/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/PropertyData.cs
+++ b/Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/PropertyData.cs
@@ -66,7 +66,7 @@ namespace EdFi.Ods.CodeGen.Generators.Resources
         {
             get
             {
-                return !Property.PropertyType.IsNullable && !Property.IsServerAssigned && CSharpDefaultHasDomainMeaning(Property);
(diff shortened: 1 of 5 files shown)


Utilities/CodeGeneration/EdFi.Ods.CodeGen/Generators/Resources/PropertyData.cs
```diff
@@ -346,5 +355,52 @@ public static PropertyData CreateNullProperty(ResourceProperty property)
 
             return propertyData;
         }
+
+        private static bool CSharpDefaultHasDomainMeaning(ResourceProperty property)
```
gmcelhanon, line 359: Rather than copy this method here, I think we should look at converting this into an extension method on the `PropertyTypeExtensions` class. Also, then, remove the one in the _Entities.cs_ generator class and change code to use the shared version of this function there and here.

I'm thinking of the following signature:
```csharp
public static bool CSharpDefaultHasDomainMeaning(this PropertyType propertyType, string propertyName)
{
    ...
}
```

gmcelhanon (changes requested):
Just some minor changes. Looks pretty good.

PR #767, opened 2023-07-03
[ODS-5907] PUT issue update for a resource that does not support the primary key updates - API doesn’t show error message

mjaksn:


semalaiappan (approved):

semalaiappan:
@mjaksn  All commits should have signed which is missing . Please fix it . Thanks

mjaksn:
> @mjaksn All commits should have signed which is missing . Please fix it . Thanks

@semalaiappan That should be fixed now

PR #768, opened 2023-07-05
[ODS-5891] Fix issue with single ods instance and single context route

simpat-adam:


Application/EdFi.Ods.Api/Middleware/OdsInstanceSelector.cs
```diff
@@ -45,7 +49,7 @@ public async Task<OdsInstanceConfiguration> GetOdsInstanceAsync(IReadOnlyDiction
             throw new ApiSecurityConfigurationException("The API client has not been associated with an ODS instance.");
         }
 
-        if (apiClientContext.OdsInstanceIds.Count == 1)
+        if (apiClientContext.OdsInstanceIds.Count == 1 && !_odsRoutesRootTemplateProvider.HasOdsContextRouteTemplate)
```
gmcelhanon, line 52: I would suggest introducing a private `Lazy<bool>` field on this class and initializing it in the constructor using a value factory delegate initialization that uses the result of the `odsRouteRootTemplateProvider.GetOdsRouteTemplate()` method. Then this logic would just check the `Value` of that field (reusing the result for the _extremely_ high volume of executions of this code).

Application/EdFi.Ods.Tests/EdFi.Ods.Api/Middleware/OdsInstanceSelectorTests.cs
```diff
@@ -29,7 +31,8 @@ public void SetUp()
         {
             _apiClientContextProvider = A.Fake<IApiClientContextProvider>();
             _odsInstanceConfigurationProvider = A.Fake<IOdsInstanceConfigurationProvider>();
-            _odsInstanceSelector = new OdsInstanceSelector(_apiClientContextProvider, _odsInstanceConfigurationProvider);
+            _odsRouteRootTemplateProvider = A.Fake<IOdsRouteRootTemplateProvider>();
+            _odsInstanceSelector = new OdsInstanceSelector(_apiClientContextProvider, _odsInstanceConfigurationProvider, _odsRouteRootTemplateProvider);
```
gmcelhanon, line 35: More than just modifying this to get it to compile, it would be good to actually add some unit test coverage for this previously unidentified scenario.

gmcelhanon (changes requested):

gmcelhanon:
@simpat-adam Also, change the title of the PR to include the JIRA ticket number as a prefix.

simpat-jesus (changes requested):
Add tests similar to GetOdsInstanceAsync_ReturnsOdsInstanceConfiguration_WhenApiClientContextHasMoreThanOneOdsInstanceId_AndOneMatchingAllContextValues to cover this scenario

simpat-jesus (approved):

gmcelhanon (approved):

PR #795, opened 2023-07-26
[ODS-5922] KeyChanges and Deletes endpoints should return 404 instead of 500 when resource does not exist

simpat-adam:


Postman Test Suite/Ed-Fi ODS-API ChangeQueries Key Changes and Deletes Test Suite.postman_collection.json
```diff
@@ -10346,15 +10346,9 @@
...
-													"pm.test(\"Response is 400 or 404\", () => {\r",
-													"  pm.expect(pm.response.code).to.be.oneOf([400, 404])\r",
-													"});\r",
-													"\r",
-													"if (pm.response.code == 400) {\r",
-													"  pm.test(\"Should indicate that Parent keys cannot be changed.\", () => {\r",
-													"    pm.expect(pm.response.json().message).to.include(\"Key values for this resource cannot be changed.\");\r",
-													"  });\r",
-													"}"
+													"pm.test(\"Response is 404\", () => {\r",
```
gmcelhanon, line 10349: The scenario we're resolving with this ticket actually has nothing to do specifically with Parent -- it was just noticed here because Parent got renamed to Contact and no longer exists in the model for v5.0.0. Since it still exists in v4.0.0 and we need to continue to provide support for testing previous versions of the model, we need to leave this test as-is. Also, this test, and others for Parent, are slated to be updated to support either Parent or Contact by [ODS-5921](https://tracker.ed-fi.org/browse/ODS-5921).

So, you should revert this change to the Postman collection, and then add a _new_ folder (e.g. "Non-Existing Resources") and add requests to cover this specific scenario which is this:

> When a request is made to get keyChanges or deletes for a non-existing resource, the API should respond with a 404 status code.

So, for example, the test should consist of two GET requests as follows, expecting a 404 status:
- `{{ApiBaseUrl}}/data/v3/ed-fi/nonExisting/keyChanges`
- `{{ApiBaseUrl}}/data/v3/ed-fi/nonExisting/deletes`

gmcelhanon (changes requested):

gmcelhanon (approved):

PR #802, opened 2023-08-04
[ODS-5921] Postman test coverage for Contacts (after renamed from Parent in the standard)

mjaksn:


Postman Test Suite/Ed-Fi ODS-API ChangeQueries Key Changes and Deletes Test Suite.postman_collection.json
```diff
@@ -10101,7 +10101,7 @@
 									"response": []
 								},
 								{
-									"name": "Create Parent",
+									"name": "Create Contact",
```
semalaiappan, line 10104: change to Create Parent or Contact   So that easy to understand if you are running 4.0.0 or 5.0.0

semalaiappan (changes requested):
All places change to Parent or Contact So that easy to understand if you are running 4.0.0 or 5.0.0

semalaiappan (approved):

PR #815, opened 2023-08-24
[ODS-5972] Dependency metadata transformations for authorization concerns do not support Contact in Ed-Fi Standard v5

mjaksn:
This updates `PersonResourceLoadGraphTransformer` so that it supports the renaming of Parent to Contact in Ed-Fi Standard v5 in addition to supporting data standard v4.


semalaiappan:
image

semalaiappan (approved):

PR #822, opened 2023-09-06
[ODS-5920] Finalize approach for handling both Parent and Contact in Integration Tests

mjaksn:


semalaiappan (changes requested):
IntegrationTests Projects are verified and looks good .  Still few other projects changes required
image

mjaksn:
The remaining projects are now updated. The only place I left separate DS v5 and DS v4 versions is for `AggregateDependencyControllerTests` in `EdFi.Ods.WebApi.IntegrationTests` since it is a single approval test, and the contents of the "approved" file will always be specific to the data standard in use.

semalaiappan (approved):

PR #825, opened 2023-09-13
[ODS-5743] Unit test is incorrectly passing in OpenApiMetadataDefinitionsFactoryTests 

mjaksn:
This updates two unit tests to align with changes elsewhere in the solution and function as initially intended.

### DIFF
--- a/Application/Test.Common/AssertHelper.cs
+++ b/Application/Test.Common/AssertHelper.cs
@@ -19,6 +19,11 @@ namespace Test.Common
         /// <param name="asserts">Asserts to invoke</param>
         public static void All(params Action[] asserts)
         {
+            if (asserts == null || asserts.Length == 0)
+            {
+                Assert.Fail($"No asserts were provided for invocation");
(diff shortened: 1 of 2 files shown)


Application/Test.Common/AssertHelper.cs
```diff
@@ -19,6 +19,11 @@ public static class AssertHelper
         /// <param name="asserts">Asserts to invoke</param>
         public static void All(params Action[] asserts)
         {
+            if (asserts == null || asserts.Length == 0)
+            {
+                Assert.Fail($"No asserts were provided for invocation");
+            }
```
mjaksn, line 25: This change prevents unit tests from passing when they do not invoke any assertions. If a unit test needs to check for the non-existence of something, this wouldn't be the appropriate helper to use. Therefore, if this method is passed an empty array it is an indication that something has not worked as intended in the test calling it.

simpat-adam (approved):

PR #840, opened 2023-10-03
[ODS-5991] Improve error message

simpat-adam:


Postman Test Suite/Ed-Fi ODS-API Integration Test Suite.postman_collection.json
```diff
@@ -6608,7 +6608,7 @@
 													"    });\r",
 													"\r",
 													"    if (request && request.FiscalYear) {\r",
```
semalaiappan, line 6610: Existing if condtion is not working at all .  So I updated here with one test case here . Please use the same format other places also . you  can debug with current code to see whether existing code if condition or not also 
pm.test("Should return a message indicating that FiscalYear is required", () => {

    const responseItem = pm.response.json();
    const expectedErrorMessage1 = "The request is invalid.";
    pm.expect(responseItem.message).to.equal(expectedErrorMessage1);

    var requestItem = JSON.parse(pm.request.body.raw);
    if (requestItem && requestItem.functionDimensionReference && requestItem.functionDimensionReference.fiscalYear !== undefined) {
        //console.log(requestItem.functionDimensionReference.fiscalYear);
        //console.log("ModelState:", responseItem.modelState["request.FiscalYear"][0]); 
        pm.expect(responseItem.modelState["request.FiscalYear"][0]).to.equal("Required");
    }
    
});

simpat-adam: I decided to remove the if statements because I can't figure out what their purpose was. I could be wrong but I don't think they were needed in the first place. We really just need to test for the proper error message every time.

semalaiappan (approved):

PR #878, opened 2023-11-08
[ODS-6077] Check Ed-Fi-ODS-Implementation Repo PR exists or not step Added

semalaiappan:


.github/workflows/Trgr InitDev workflows in Implementation repo.yml
```diff
@@ -57,8 +57,29 @@ jobs:
           $current_branch = '${{env.HEAD_REF}}'
         }
         Write-Host "Current Branch: $current_branch"   
-        echo "current_branch=$current_branch" >> $Env:GITHUB_ENV                   
+        echo "current_branch=$current_branch" >> $Env:GITHUB_ENV
+    - name: Check Ed-Fi-ODS-Implementation Repo PR exists or not
+      shell: pwsh
+      run: |
+        $accessToken = "${{ env.EDFI_ODS_IMP_TOKEN  }}"
+        $apiUrl = "https://api.github.com/repos/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pulls?state=open&head=Ed-Fi-Alliance-OSS:${{env.current_branch}}"
```
stephenfuqua, line 65: Can we use variable substitution to inject the name of the name of the organization and the repository, so that this workflow can run in a fork?

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
```diff
@@ -57,8 +57,29 @@ jobs:
...
+        $headers = @{
+            Authorization = "Bearer $accessToken"
+            Accept = "application/vnd.github.v3+json"
+        }
+        $prList = Invoke-RestMethod -Uri $apiUrl -Headers $headers
```
stephenfuqua, line 70: Can you also capture and print the rate limit headers, which might be useful for diagnostics? See https://docs.github.com/en/rest/overview/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
```diff
@@ -57,8 +57,34 @@ jobs:
...
+        $rateLimitRemaining = $response.Headers['x-ratelimit-remaining']
+        Write-Host "x-ratelimit-limit: $rateLimitLimit"
+        Write-Host "x-ratelimit-remaining: $rateLimitRemaining"              
+        foreach ($pr in $response) {
+         Write-Host "current branch: ${{env.current_branch}}"
+         Write-Host "ref: $($pr.head.ref)"
+          if ($pr.head.ref -eq "${{ env.current_branch }}") 
+          {
+            Write-Host "PR #$($pr.number): $($pr.head.label)"
+            echo "EXIT_STEP=true" >> $Env:GITHUB_ENV
```
stephenfuqua, line 83: I think that I follow the logic here. If we find a match, then `EXIT_STEP == true` and the steps below will run. If no branch match found, then the steps won't run.

Suggestion: write a [formatted error message](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message) after the foreach loop if `EXIT_STEP != true`. This will make it easier to identify why the problem occurred. It creates an annotation that is very visible in the action results, and the message will be formatted in red in the log.

Final suggestion: `echo` is okay here, but it is a little strange to me seeing `Write-Host` above and then `echo` here. Wouldn't it be better for this to remain consistent?

stephenfuqua:
@semalaiappan please create forks of Ed-Fi-ODS and Ed-Fi-ODS-Implementation to demonstrate this process.

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
```diff
@@ -57,8 +57,44 @@ jobs:
...
+            if ($pr.head.ref -eq "${{ env.current_branch }}") 
+            {
+                Write-Host "PR #$($pr.number): $($pr.head.label)"
+                echo "EXIT_STEP=true">> $env:GITHUB_ENV
+            }
+        }
+    - name: Check for EXIT_STEP Status
+      if: env.EXIT_STEP == 'true'
+      run: | 
+           echo "::notice::Ed-Fi-ODS-Implementation already have same branch ${{ env.current_branch }} running all Initdev builds ,so Skipping this build"
```
stephenfuqua, line 95: Oh, I completely misunderstood the intent of this pull request before! But now I get it after reading note. I still find it strange that you have used `echo` here instead of `Write-Host` as seen elsewhere. This is good and makes a lot of sense.

Another thing I would like to see is in the remote job in `Ed-Fi-ODS-Implementation` is that the job quits if the repo doesn't have the required branch, instead of retrying over and over. We can do that in another ticket.

semalaiappan: why I used echo for setting environment variable in PowerShell  echo "EXIT_STEP=true">> $env:GITHUB_ENV https://www.jamescroft.co.uk/setting-github-actions-environment-variables-in-powershell/

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
semalaiappan: why I used echo for setting environment variable in PowerShell echo "EXIT_STEP=true">> $env:GITHUB_ENV https://www.jamescroft.co.uk/setting-github-actions-environment-variables-in-powershell/

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
semalaiappan: And also I used in many other builds as well

stephenfuqua: In PowerShell, `echo` is just an alias of [Write-Output](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-output?view=powershell-7.3). Also, `Write-Host` and `Write-Output` are very similar; after many years of using them, I still haven't fully understood why they both exist. I'm not going to stop the PR over this topic, but our official coding standard for PowerShell is to write out the commands in their full PowerShell form, not in shortcut form like `echo` (another example: using `Remove-Item` instead of `rm`).

stephenfuqua (approved):

PR #900, opened 2023-12-28
[ODS-6088]Cancel a running jobs on new commits

semalaiappan:


.github/workflows/Cleanup Caches by a branch.yml
```diff
@@ -8,7 +8,11 @@ on:
   pull_request:
     branches: [main, 'ODS-*']
     types: [ closed ]
-  
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
simpat-jesus, line 13: This workflow executes only when a PR is closed, so there will not be an in progress workflow to cancel. This change can be reverted.

simpat-jesus (changes requested):

.github/workflows/Cleanup Caches by a branch.yml
semalaiappan: Revert it back . Thanks

vimayya (approved):

PR #903, opened 2024-01-02
[ODS-6139] Use Problem Details (RFC 9457) for reporting errors and validation errors

gmcelhanon:


Application/EdFi.Ods.Api/Attributes/NoDangerousTextAttribute.cs
```diff
@@ -22,7 +22,8 @@ protected override ValidationResult IsValid(object value, ValidationContext vali
             }
 
             return CrossSiteScriptingValidation.IsDangerousString(s, out int matchIndex)
-                ? new ValidationResult($"{validationContext.DisplayName} contains a potentially dangerous value.")
+                ? new ValidationResult($"{validationContext.DisplayName} contains a value that could be dangerous for downstream systems using this data. Try to avoid the use of special symbols like '<', '>' or '&' without surrounding spaces.",
```
gmcelhanon, line 25: Data validation errors are to target end-users of the source applications (e.g. SIS software). Making this message more user-friendly.

Application/EdFi.Ods.Api/Container/Modules/ApplicationModule.cs
```diff
@@ -497,6 +490,14 @@ void RegisterMiddleware()
                 builder.RegisterType<OdsInstanceSelector>()
                     .As<IOdsInstanceSelector>()
                     .SingleInstance();
+
+                builder.RegisterType<ErrorTranslator>().SingleInstance();
+                builder.RegisterType<ModelStateKeyConverter>().EnableClassInterceptors().SingleInstance();
```
gmcelhanon, line 495: Converts ModelState keys to JSONPaths (after validation).

Application/EdFi.Ods.Api/ExceptionHandling/ModelStateKeyConverter.cs
```diff
@@ -0,0 +1,111 @@
...
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using Autofac.Extras.DynamicProxy;
+using EdFi.Ods.Common.Models.Resource;
+
+namespace EdFi.Ods.Api.ExceptionHandling;
+
+[Intercept("cache-model-state-key")]
```
gmcelhanon, line 13: An interceptor has been added here to prevent the same path from ever being translated twice. This may or may not be a good optimization, but the idea is to reduce the number of string allocations that would be caused by repeatedly translating the same model paths to JSON Paths.

Application/EdFi.Ods.Common/Exceptions/EdFiProblemDetailsExceptionBase.cs
```diff
@@ -0,0 +1,131 @@
...
+    public abstract string Title { get; }
+
+    /// <inheritdoc cref="IEdFiProblemDetails.Status" />
+    public abstract int Status { get; }
+
+    /// <inheritdoc cref="IEdFiProblemDetails.CorrelationId" />
+    public string CorrelationId { get; set; }
+
+    /// <inheritdoc cref="IEdFiProblemDetails.ValidationErrors" />
+    Dictionary<string, string[]> IEdFiProblemDetails.ValidationErrors { get; set; }
```
gmcelhanon, line 122: `ValidationErrors` and `Errors` are implemented explicitly to prevent them from being logged with the exception. Their contents are intended for the API client, and their end user(s).

Application/EdFi.Ods.Common/Models/Resource/ResourceClassBase.cs
```diff
@@ -269,7 +269,7 @@ public string PluralName
         /// <summary>
         /// Gets the identifying properties of the resource that are introduced in the local context, including those in references, as necessary.
         /// </summary>
-        public IReadOnlyList<ResourceProperty> IdentifyingProperties => AllProperties
+        public virtual IReadOnlyList<ResourceProperty> IdentifyingProperties => AllProperties
```
gmcelhanon, line 272: These members were made virtual to support unit tests which were too difficult to setup without using mocks.

Application/EdFi.Ods.Common/Validation/ValidationAttributeStore.cs
```diff
@@ -0,0 +1,285 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+
+namespace EdFi.Ods.Common.Validation
```
gmcelhanon, line 17: This is copied from .NET 8 because modifications are needed to achieve the validation behavior we need (full rather than short-circuited validation, and incorporating Profiles restrictions).

Application/EdFi.Ods.Common/Validation/Validator.cs
```diff
@@ -0,0 +1,692 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using EdFi.Ods.Common.Extensions;
+using EdFi.Ods.Common.Models;
+using EdFi.Ods.Common.Models.Domain;
+using EdFi.Ods.Common.Models.Resource;
+
+namespace EdFi.Ods.Common.Validation
```
gmcelhanon, line 19: This is copied from .NET 8 because modifications are needed to achieve the validation behavior we need (full rather than short-circuited validation, and incorporating Profiles restrictions).

Application/EdFi.Ods.Common/Validation/Validator.cs
```diff
@@ -0,0 +1,692 @@
...
+        /// <returns>
+        ///     A set of key value pairs, where the key is a validation context for the property and the value is its current
+        ///     value.
+        /// </returns>
+        /// <remarks>Ignores indexed properties.</remarks>
+        // [RequiresUnreferencedCode(ValidationContext.InstanceTypeNotStaticallyDiscovered)]
+        private static List<KeyValuePair<ValidationContext, object?>> GetPropertyValues(object instance,
+            ValidationContext validationContext)
+        {
+            // ==========================
```
gmcelhanon, line 534: All customizations made to the code is commented with the text: `// NOTE: Customization`

Application/EdFi.Ods.Common/Validation/ValidationAttributeStore.cs
simpat-jesus: Do we need to remove this as part of the migration to .NET 8?

Application/EdFi.Ods.Common/Validation/Validator.cs
simpat-jesus: Do we need to remove this as part of the migration to .NET 8?

Application/EdFi.Ods.Common/Validation/ValidationAttributeStore.cs
gmcelhanon: Yes, probably would be good because it is an unmodified supporting component to the Validator.cs class, which is modified for our needs.

Application/EdFi.Ods.Common/Validation/Validator.cs
gmcelhanon: This one needs to stay. It's a version of Validator.cs that is customized specifically for our needs with Profiles and to achieve the behavior that we're looking for for full object graph validation.

Application/EdFi.Ods.Api/Models/RESTError.cs
```diff
@@ -26,7 +26,7 @@ public class RESTError
         public string CorrelationId { get; set; }
 
         // See ErrorTranslator.GetErrorMessage
-        [DataMember(Name = "modelState")]
-        public Dictionary<string, string[]> ModelState { get; set; }
+        [DataMember(Name = "validationErrors")]
+        public Dictionary<string, string[]> ValidationErrors { get; set; }
```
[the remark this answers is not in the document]
gmcelhanon: This class is obsolete and I thought it had been removed. I'll remove it.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/OwnershipBased/OwnershipBasedAuthorizationContextData.cs
```diff
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace EdFi.Ods.Api.Security.AuthorizationStrategies.OwnershipBased
+﻿namespace EdFi.Ods.Api.Security.AuthorizationStrategies.OwnershipBased
```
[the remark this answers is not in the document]
gmcelhanon: I've added it.

Application/EdFi.Ods.Common/Attributes/DescriptorExistsAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+    public DescriptorExistsAttribute(string descriptorName)
+    {
+        _descriptorName = descriptorName;
+    }
+
+    protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+    {
+        if (value == null)
+        {
+            return ValidationResult.Success;
```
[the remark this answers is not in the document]
gmcelhanon: No, because this isn't a "Required" type of check... it's checking whether the descriptor _value_ provided actually exists in the ODS (using the local cache). If there is no value provided, there's nothing to check -- we return Success. To reduce confusion, I could perhaps rename the attribute to `DescriptorValueExists` (this would require code gen changes).

Application/EdFi.Ods.Api/Security/Authorization/PersonResourceGraphAuthorizationTransformer.cs
```diff
@@ -141,6 +159,17 @@ private static void ApplyContactTransformation(BidirectionalGraph<Resource, Asso
                 resourceGraph.RemoveEdge(directContactDependency);
                 resourceGraph.AddEdge(new AssociationViewEdge(studentContactAssociationResource, directContactDependency.Target, directContactDependency.AssociationView));
             }
+
+            string LogAndThrowException()
```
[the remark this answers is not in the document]
gmcelhanon: This was separated out into separate method not for reuse, but because the containing structure uses pattern matching and you can only provide an expression here (not a block). Because I wanted to expand what was being done, I was forced to move it to a separate method.

semalaiappan (approved):

PR #929, opened 2024-01-25
[ODS-6111] Update EdFi.Security.DataAccess to .Net 8

JBrenesSimpat:


.github/workflows/edFi.security.dataaccess pullrequest.yml
```diff
@@ -2,10 +2,15 @@ name: EdFi.Security.DataAccess Pull request build and test
 
 on:
   pull_request:
-    branches: [main]
+    branches:
+      - main-6x
+    paths:
+      - Application/EdFi.Security.DataAccess/**/*
+      - test/EdFi.Security.DataAccess.UnitTests/**/*
```
stephenfuqua, line 9: Should be `tests` with 's' on the end

.github/workflows/edFi.security.dataaccess manual.yml
stephenfuqua, line 1: Please rename as `Lib edFi.security.dataaccess manual.yml`.

.github/workflows/edFi.security.dataaccess pullrequest.yml
stephenfuqua, line 1: Please rename as `Lib edFi.security.dataaccess pullrequest.yml`.

Application/EdFi.Security.DataAccess/Repositories/CachedSecurityRepository.cs
stephenfuqua, line 1: Please restore this file. It is still used in ODS/API 6.x code (see `EdFi.Ods.Api.Container.Modules.SecurityRepositoryModule`). In ODS/API 7.x there were some recent changes to improve or remove some of the caching, which probably explains why this file would have been removed from the `main` branch.

Application/EdFi.Security.DataAccess/Repositories/ISecurityRepository.cs
```diff
@@ -4,35 +4,39 @@
...
 using EdFi.Security.DataAccess.Models;
 
-namespace EdFi.Security.DataAccess.Repositories
-{
-    public interface ISecurityRepository
-    {
-        Action GetActionByHttpVerb(string httpVerb);
+namespace EdFi.Security.DataAccess.Repositories;
 
-        Action GetActionByName(string actionName);
```
stephenfuqua, line 15: Please restore `GetActionByHttpVerb`. If we change this interface we will break other code that depends on it.

Application/EdFi.Security.DataAccess/Repositories/ISecurityRepository.cs
```diff
@@ -4,35 +4,39 @@
...
+/// <summary>
+/// Defines higher-level methods for obtaining security metadata for making authorization decisions.
+/// </summary>
+/// <remarks>Implementations of this interface must be configured with a named <see cref="IInterceptor" /> registration of "cache-security".</remarks>
+[Intercept("cache-security")]
```
stephenfuqua, line 17: I'm not sure what this `[Intercept(...)]` does. Are you sure it is necessary for the .NET 8 upgrade?

Application/EdFi.Security.DataAccess/Repositories/ISecurityRepository.cs
```diff
@@ -4,35 +4,39 @@
...
+public interface ISecurityRepository
+{
+    Action GetActionByName(string actionName);
 
-        AuthorizationStrategy GetAuthorizationStrategyByName(string authorizationStrategyName);
+    AuthorizationStrategy GetAuthorizationStrategyByName(string authorizationStrategyName);
 
-        IEnumerable<ClaimSetResourceClaimAction> GetClaimsForClaimSet(string claimSetName);
+    IList<ClaimSetResourceClaimAction> GetClaimsForClaimSet(string claimSetName);
```
stephenfuqua, line 24: Changing `IEnumerable` to `IList`: This is probably safe to do, but I doubt that this change is necessary for the 6.x codebase. Best to keep the original interface though, and change back to `IEnumerable`, unless you have a .NET 8 warning or code suggestion. If you do have such a warning, then `IReadOnlyList` is probably a better choice: based on the names, this code is only for retrieving data, not for changing it, so a read only list would be appropriate. I also looked at `ClaimsIdentityProvider` in `EdFi.Ods.Api` and confirmed that `IReadOnlyList` will work there.

Application/EdFi.Security.DataAccess/Repositories/SecurityRepository.cs
```diff
@@ -5,119 +5,129 @@
...
 namespace EdFi.Security.DataAccess.Repositories
 {
-    public class SecurityRepository : SecurityRepositoryBase, ISecurityRepository
+    /// <summary>
+    /// Provides higher level access to raw security metadata.
+    /// </summary>
+    public class SecurityRepository : ISecurityRepository
     {
-        private readonly ISecurityContextFactory _securityContextFactory;
+        private readonly ISecurityTableGateway _securityTableGateway;
```
stephenfuqua, line 19: Please revert this change, as it is likely to break the applications that are expecting to inject `ISecurityContextFactory`.

Application/EdFi.Security.DataAccess/Repositories/SecurityRepository.cs
```diff
@@ -5,119 +5,129 @@
...
         }
 
-        private Application GetApplication()
+        public virtual Action GetActionByName(string actionName)
         {
-            using var context = _securityContextFactory.CreateContext();
-
-            return context.Applications.First(
-                app => app.ApplicationName.Equals("Ed-Fi ODS API", StringComparison.InvariantCultureIgnoreCase));
+            return _securityTableGateway.GetActions().First(a => a.ActionName.Equals(actionName, StringComparison.OrdinalIgnoreCase));
```
stephenfuqua, line 28: The hard-coded "Ed-Fi ODS API" looks strange here. I don't like it. But we need to keep it, because other applications that depend on this library expect that.

Application/EdFi.Security.DataAccess/Utils/ValidateDatabase.cs
stephenfuqua, line 1: I do not see this file in use by Admin App or the ODS/API, so this removal may be safe.

tests/EdFi.Security.DataAccess.UnitTests/Repositories/CachedSecurityRepositoryTests.cs
stephenfuqua, line 1: Please remember to restore this when you restore `CachedSecurityRepository`.

stephenfuqua (commented):
With this ODS/API 6.2 update to .NET 8, please keep in mind that we do not want to copy any changes from `main` except for those changes that are specifically related to .NET 8. 

In some circumstances we can also apply small C# improvements so long as we are 100% sure they do not change the way the code works.

stephenfuqua (approved):

PR #933, opened 2024-01-26
[ODS-6175] Migrate GenerateSecurityGraphs to .NET 8

JBrenesSimpat:


Utilities/GenerateSecurityGraphs/GenerateSecurityGraphs/Program.cs
```diff
@@ -99,18 +99,19 @@ private static async Task Main(string[] args)
...
 
             string connectionString;
-
+            
             if (options.User == null || options.Password == null)
             {
                 connectionString = string.Format(
-                    "Server={0};Database={1};Trusted_Connection=True",
+
+                    "Server={0};Database={1};Trusted_Connection=True;Encrypt=False",
```
stephenfuqua, line 107: I need to learn more about how this connection string is used. I don't like having a hard-coded connection string, and turning off encryption here is a bad idea. I understand why we do that for localhost development, but if this ever runs on a real server then this is bad.

@vimayya please weigh in on how we should handle this connection string. Should we perhaps add a new command line parameter for encryption? Or change the command line parameter to require that the user provide the connection string?

vimayya: I would go for full connection string passed as parameter. It would be a breaking change but we could go that route considering this utility is not likely to have additional integrations.

stephenfuqua (approved):

PR #938, opened 2024-01-30
[ODS-6171] Migrate SdkGen to .NET 8

JBrenesSimpat:


Utilities/SdkGen/EdFi.SdkGen.Console/Options.cs
```diff
@@ -28,7 +28,7 @@ public sealed class Options
         [Option(
             'v',
             "cliVersion",
-            Default = "6.0.1",
+            Default = "6.2.0",
             HelpText = "the version of openapi-codegen-cli to download")]
         public string CliVersion { get; set; }
```
stephenfuqua, line 33: Doesn't this refer to https://github.com/OpenAPITools/openapi-generator instead of our software? In which case, maybe we should set the default to the newest one available?

JBrenesSimpat: Updated to 7.2.0 to support net8.0

stephenfuqua (approved):

PR #940, opened 2024-02-02
[ODS-6167] Migrate Ed-Fi-Ods solution to .NET 8

JBrenesSimpat:


Application/EdFi.Ods.Api/ScheduledJobs/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,11 +8,13 @@
 using log4net;
 using Microsoft.Extensions.DependencyInjection;
 using Quartz;
+using Quartz.AspNetCore;
 
 namespace EdFi.Ods.Api.ScheduledJobs.Extensions
 {
     public static class ServiceCollectionExtensions
     {
+        [System.Obsolete]
```
stephenfuqua, line 17: Why is this marked as `obsolete`? I do not see any reason to change that during this .NET 8 upgrade.

Application/EdFi.Ods.Api/Startup/OdsStartupBase.cs
```diff
@@ -367,9 +365,7 @@ void SetStaticResolvers()
                             .GetResourceModel());
 
                 EntityExtensionsFactory.Instance = Container.Resolve<IEntityExtensionsFactory>();
-
-                DbConfiguration.SetConfiguration(new DatabaseEngineDbConfiguration(Container.Resolve<DatabaseEngine>()));
```
stephenfuqua, line 371: What replaces this line of code? Is changing this really necessary for the .NET 8 upgrade?

Application/EdFi.Ods.Tests/EdFi.Ods.Api/Caching/StubSecurityContext.cs
```diff
@@ -23,21 +20,21 @@ public static ISecurityContext GetMockedSecurityContext()
         {
             var securityContext = A.Fake<ISecurityContext>();
             // The underlying SecurityRepository implementation expects this application, so force it to be there in the fake
```
stephenfuqua, line 22: This comment no longer seems relevant. Perhaps it would be good to remove it.

I did not think about the following idea when you were working on the similar class in another project. It might be good to add a comment explaining what this class is doing. Maybe something like this:

```csharp
// This stub class allows us to compile and perform some tests in repositories
// that rely on `ISecurityContext`. However, `DbSet` is very difficult to work with
// in unit tests, and we no longer try. This stub class returns `null` for everything,
// under the assumption that test classes that use this stub are only testing 
// situations where the properties below will not be called.
```

Now, looking at this again, I also realize that we could have gotten rid of `GetFakeDbSet` altogether and instead simply set each of these properties `= null`!

Application/EdFi.Ods.Tests/EdFi.Ods.Sandbox/Repositories/ClientAppRepoTests.cs
```diff
@@ -23,30 +25,38 @@ public class When_calling_the_clientAppRepo
...
                 {
                     Name = "ClientAppRepoTest" + Guid.NewGuid()
                         .ToString("N")
                 };
 
-                var dbSet = new TestDbSet<ApiClient> {_testClient};
-
-                A.CallTo(() => usersContext.Clients)
-                    .Returns(dbSet);
+                _clientAppRepo.UpdateClient(_testClient);
```
stephenfuqua, line 59: You've turned this from a unit test to an integration test. At first I was going to ask you to delete this class altogether, since unit testing this repository is impractical.

However, after scrolling down some more, I remember now that we have a separate `EdFi.Admin.DataAccess.IntegrationTests` project. The `ClientAppRepo` class is part of `EdFi.Admin.DataAccess`, so this unit test never should have been under the `EdFi.Ods.Sandbox` in the first place.

Please keep your changes but move the file into project `EdFi.Admin.DataAccess.IntegrationTests`.

tests/EdFi.Security.DataAccess.UnitTests/EdFi.Security.DataAccess.UnitTests.csproj
```diff
@@ -18,7 +18,7 @@
   <ItemGroup>
 	  <PackageReference Include="FakeItEasy" Version="8.1.0" />
 	  <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
-	  <PackageReference Include="NUnit" Version="4.0.1" />
+	  <PackageReference Include="NUnit" Version="3.13.3" />
```
stephenfuqua, line 21: Please switch back to NUnit 4 unless there is some reason we cannot use it. If we cannot use it, then please pin the version so that it is impossible to update it, for example with `Version="[3.13.0,4.0.0)"`.

tests/EdFi.TestFixture/DbContextExtensions.cs
```diff
@@ -23,16 +22,11 @@ public static void CatchAndWriteDbValidationExceptions(Action action)
             {
                 action();
             }
-            catch (DbEntityValidationException exception)
+            catch (ValidationException exception)
             {
-                var errorTexts =
-                    exception.EntityValidationErrors.SelectMany(x => x.ValidationErrors)
-                        .Select(x => x.ErrorMessage);
+                var errorTexts = exception.ValidationResult.ErrorMessage;
```
stephenfuqua, line 27: Is this no longer an array? If that is the case, it may make sense to rename the variable to `errorText` (singular), or just get rid of the variable altogether and simply use `Console.WriteLine(exception.ValidationResult.ErrorMessage);`.

tests/EdFi.TestFixture/EdFi.TestFixture.csproj
```diff
@@ -13,7 +13,9 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="FakeItEasy" Version="8.1.0" />
-    <PackageReference Include="NUnit" Version="4.0.1" />
-    <PackageReference Include="EntityFramework" Version="6.4.4" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Abstractions" Version="8.0.1" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="8.0.1" />
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.0" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
```
stephenfuqua, line 19: Please switch back to NUnit 4 if possible.

PR #942, opened 2024-02-03
[ODS-6152] Migrate DataLoading solution to .NET 8

axelmarquezh:
I've also updated the related libraries in our artifacts repository:
- EdFi.Suite3.LoadTools
- EdFi.Suite3.BulkLoadClient.Console


simpat-jesus (changes requested):
Some dependencies also need to be upgraded. Please use the [assistant](https://dotnet.microsoft.com/en-us/platform/upgrade-assistant) to handle those dependencies.

axelmarquezh:
Thanks Jesus. I've updated the dependencies as well.
Note that I also removed the `EdFi.XmlLookup.Console` project, which seems to have been missed while developing ODS-6084.

simpat-jesus (approved):

PR #945, opened 2024-02-05
[ODS-6250] Update to Virtual Models for Admin.DataAccess

JBrenesSimpat:
Several changes are included to enable Web-API Sandbox and Swagger execution with MSSQL and Postgres.


Application/EdFi.Admin.DataAccess/Contexts/IUsersContext.cs
```diff
@@ -16,7 +16,7 @@ public interface IUsersContext : IDisposable
     {
         DbSet<User> Users { get; set; }
 
-        DbSet<ApiClient> Clients { get; set; }
+        DbSet<ApiClient> ApiClients { get; set; }
```
stephenfuqua, line 19: Why change this? This will also cause breaking changes in the upcoming Admin API 1 and Admin App releases. I do agree that `ApiClients` is a _better_ name, but I'm hesitant about causing us extra work by changing this interface.

Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs
```diff
@@ -35,7 +35,7 @@ public static string UserIdColumn
 
         public DbSet<User> Users { get; set; }
 
-        public DbSet<ApiClient> Clients { get; set; }
+        public virtual DbSet<ApiClient> ApiClients { get; set; }
```
stephenfuqua, line 38: Why did these need to be made virtual?

Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs
```diff
@@ -74,6 +74,11 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                     r =>
                         r.HasOne<ApiClient>().WithMany().HasForeignKey("ApiClientId"));
 
+            modelBuilder.Entity<Application>()
+                .HasMany(a => a.Profiles)
+                .WithMany(a => a.Applications)
+                .UsingEntity("ProfileApplications");
```
stephenfuqua, line 80: This seems like a good change, but why? Is this addressing a bug fix?

stephenfuqua (changes requested):
We need to be very careful to avoid breaking changes. I do not see that any of these changes are necessary for the .NET 8 upgrade. If any are related to bug fixes that have been applied in the ODS/API 7.x code, then we should wait until Vinaya gives us a bug fix ticket. In some cases we may not want to fix a bug in 6.x because it would cause unexpected behavior.

Application/EdFi.Admin.DataAccess/Contexts/IUsersContext.cs
```diff
@@ -37,7 +37,7 @@ public interface IUsersContext : IDisposable
         DbSet<OwnershipToken> OwnershipTokens { get; set; }
 
         DbSet<ApiClientOwnershipToken> ApiClientOwnershipTokens { get; set; }
-
+
```
stephenfuqua, line 40: This commit introduces a lot of extra space characters.

stephenfuqua:
CodeQL and EdFi.Common failures are acceptable in this pull request.

Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs
```diff
@@ -74,25 +74,12 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                     r =>
                         r.HasOne<ApiClient>().WithMany().HasForeignKey("ApiClientId"));
 
-            modelBuilder.Entity<Application>()
-                .HasMany(a => a.Profiles)
-                .WithMany(a => a.Applications)
-                .UsingEntity("ProfileApplications");
```
stephenfuqua: The `main` branch [still has this](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/main/Application/EdFi.Admin.DataAccess/Contexts/UsersContext.cs). This looks like valuable code. Are you 100% sure we don't need it?

stephenfuqua (approved):

PR #949, opened 2024-02-07
[ODS-5936] API Publisher sometimes converts dates to timestamps

axelmarquezh:


simpat-jesus (approved):
Looks good, please update the branch

PR #988, opened 2024-02-29
[ODS-6283] Refine problem details responses for 401, 404, 405 status codes

gmcelhanon:


Application/EdFi.Ods.Api/Filters/EnforceAssignedProfileUsageFilter.cs
```diff
@@ -160,10 +153,31 @@ void SetForbiddenResponse()
...
+                                ? contentTypes.Readable
+                                : contentTypes.Writable)
+                            != null)
+                    .ToArray();
+            }
+            catch (FeatureDisabledException ex)
+            {
+                // Force a client configuration error response if profile(s) are assigned and the Profiles feature is disabled.
+                throw new ClientConfigurationException(ClientConfigurationException.DefaultDetail,
+                    $"The API client has been configured with an assigned profile but the '{ex.FeatureName}' feature is disabled.");
```
axelmarquezh, line 179: This exception will never happen (as of now), right? Because `GetAssignedProfilesForRequest()` doesn't get called if the Profiles feature is disabled [because of this condition](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/bbd005dc9cb2a85e9b3928c2fd13c3a4b14c140f/Application/EdFi.Ods.Api/Filters/EnforceAssignedProfileUsageFilter.cs#L55).

Also note this is the only place where `ClientConfigurationException` is used.

Application/EdFi.Ods.Features/ChangeQueries/Controllers/DeletesController.cs
```diff
@@ -76,13 +77,15 @@ public async Task<IActionResult> Get(string schema, string resource, [FromQuery]
             {
                 _logger.Debug("ChangeQueries is not enabled.");
 
-                return ControllerHelpers.NotFound("ChangeQueries is not enabled.");
+                return ControllerHelpers.FeatureDisabled(
+                    ChangeQueriesConstants.FeatureName,
+                    _logContextAccessor.GetCorrelationId());
```
axelmarquezh, line 82: If I disable the `ChangeQueries` feature and call the deletes/keyChanges endpoint, I get a generic `System Error` because the Deletes/KeyChangesController can't be constructed since some of its dependencies aren't registered in the DI container because the feature is disabled.

`Autofac.Core.DependencyResolutionException: None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'EdFi.Ods.Features.ChangeQueries.Controllers.DeletesController' can be invoked with the available services and parameters:
Cannot resolve parameter 'EdFi.Ods.Features.ChangeQueries.Repositories.DeletedItems.IDeletedItemsResourceDataProvider deletedItemsResourceDataProvider' of constructor ...`

axelmarquezh (changes requested):

Application/EdFi.Ods.Features/ChangeQueries/Controllers/DeletesController.cs
gmcelhanon: I've added a new virtual method on the `ConditionalModule` class: `ApplyNullObjectPatternRegistrations`, which is called when the module returns `false` from `IsSelected`. It's virtual rather than abstract, so implementation by derived classes is optional. This allows us to implement the "Null Object Pattern" for dependencies that aren't used but need to be supplied for reasons like you've identified above.

Application/EdFi.Ods.Api/Filters/EnforceAssignedProfileUsageFilter.cs
gmcelhanon: I have gone a bit further and done some refactoring around disabled feature handling generally, but for this particular change I am removing the unused exception class and the surrounding `try/catch` block. Seems like it could potentially be useful, but if that happens we can add it (or something similar) back.

axelmarquezh (approved):

gmcelhanon (commented):
All changes look good now.

mjaksn (approved):

PR #993, opened 2024-03-05
[ODS-3401] Cycles in claims taxonomy (security metadata for resource claims) causes requests to hang

mjaksn:


gmcelhanon:
@mjaksn Just wondering if you might be able to do this more directly by just switching to using a `HashSet<T>` in the `GetResourceClaimLineageForResourceClaim` method?

```csharp
        private IEnumerable<ResourceClaim> GetResourceClaimLineageForResourceClaim(string resourceClaimUri)
        {
            // ==============================
            // Use a HashSet<T> here instead of List<T>?     
            // ==============================
            var resourceClaimLineage = new List<ResourceClaim>();

            ResourceClaim resourceClaim;

            try
            {
                resourceClaim = _securityTableGateway.GetResourceClaims()
                    .SingleOrDefault(rc => rc.ClaimName.Equals(resourceClaimUri, StringComparison.OrdinalIgnoreCase));
            }
            catch (InvalidOperationException ex)
            {
                // Use InvalidOperationException wrapper with custom message over InvalidOperationException
                // thrown by Linq to communicate back to caller the problem with the configuration.
                throw new InvalidOperationException($"Multiple resource claims with a claim name of '{resourceClaimUri}' were found in the Ed-Fi API's security configuration. Authorization cannot be performed.", ex);
            }

            if (resourceClaim != null)
            {
                // ==============================
                // OLD: resourceClaimLineage.Add(resourceClaim);
                // NEW:
                if (!resourceClaimLineage.Add(resourceClaim))
                {
                    // Handle cycle problem here
                }
               // ==============================
    
                if (resourceClaim.ParentResourceClaim != null)
                {
                     resourceClaimLineage.AddRange(GetResourceClaimLineageForResourceClaim(resourceClaim.ParentResourceClaim.ClaimName));
                }
            }
            return resourceClaimLineage;
        }
```

Also, in looking at the code, I think it's an opportunity to be more efficient and declare the `HashSet<T>` in the calling method and pass it into the recursive private method, thereby avoiding the creation of multiple HashSets or Lists when this is called.

gmcelhanon (dismissed):
After direct conversation in Slack, we discussed a few change/improvements to the implementation, include:
* Switch back to `List<T>` and use a `Contains` check in lieu of a `HashSet<T>` and a check on the `Add` method because `HashSet<T>` doesn't guarantee preservation of insertion order which is important to the semantics of this method.
* Expand the resulting message to include the resource claims participating in the cycle to be included in the `errors` collection of the Problem Details response.
* Add unit test coverage since this would be difficult to test at an integration test level.

tests/EdFi.Security.DataAccess.UnitTests/Repositories/SecurityRepositoryTests.cs
```diff
@@ -165,6 +166,26 @@ public void GetResourceClaimLineage_ShouldReturnEmptyListForInvalidResourceClaim
...
+        var resourceClaimD = new ResourceClaim { ClaimName = "uri-D", ResourceClaimId = 4, ParentResourceClaim = resourceClaimB, ParentResourceClaimId = 3};
+        
+        resourceClaimA.ParentResourceClaimId = resourceClaimC.ResourceClaimId;
+        resourceClaimA.ParentResourceClaim = resourceClaimC;
+        
+        A.CallTo(() => _securityTableGateway.GetResourceClaims())
+            .Returns(new List<ResourceClaim> { resourceClaimA, resourceClaimB, resourceClaimC });
+
+        // Act & Assert
+        Should.Throw<InvalidOperationException>(() => _securityRepository.GetResourceClaimLineage("uri-C"))
```
gmcelhanon, line 186: To include the effects of the `SkipWhile` logic, I think we need to be requesting `uri-D` here, and the `securityTableGateway` mock a couple lines above should be returning all 4 resource claims.

axelmarquezh:
If I disable the Security cache, this validation fails to find the cycle because it hits the DB on each recursive call (so we get a new instance with a different hash code) so `.Contains()` always returns false.

Turning off the Security cache isn't a real-world scenario; should we care about this edge case?

```
"Security": {
    "AbsoluteExpirationMinutes": -1
}
```

Application/EdFi.Security.DataAccess/Repositories/SecurityRepository.cs
```diff
@@ -72,14 +70,21 @@ private IEnumerable<ResourceClaim> GetResourceClaimLineageForResourceClaim(strin
                 throw new InvalidOperationException($"Multiple resource claims with a claim name of '{resourceClaimUri}' were found in the Ed-Fi API's security configuration. Authorization cannot be performed.", ex);
             }
 
-            if (resourceClaim != null)
+            if (resourceClaim == null)
+            {
+                return resourceClaimLineage;
+            }
+            
+            if (resourceClaimLineage.Select(rc => rc.ClaimName).Contains(resourceClaim.ClaimName))
```
axelmarquezh, line 78: Optional: This could also be expressed as:
```suggestion
            if (resourceClaimLineage.Any(rc => rc.ClaimName == resourceClaim.ClaimName))
```

Application/EdFi.Security.DataAccess/Repositories/SecurityRepository.cs
```diff
@@ -72,14 +70,21 @@ private IEnumerable<ResourceClaim> GetResourceClaimLineageForResourceClaim(strin
...
             {
-                resourceClaimLineage.Add(resourceClaim);
+                throw new InvalidOperationException($"A cycle was detected in the resource claim hierarchy of the security metadata: '{string.Join("' -> '", resourceClaimLineage.SkipWhile(rc => rc != resourceClaim).Select(rc => rc.ClaimName))}' -> '{resourceClaimUri}'");
```
axelmarquezh, line 80: The `SkipWhile` comparison should also be based on the `ClaimName`:
```suggestion
                throw new InvalidOperationException($"A cycle was detected in the resource claim hierarchy of the security metadata: '{string.Join("' -> '", resourceClaimLineage.Select(rc => rc.ClaimName).SkipWhile(rc => rc != resourceClaim.ClaimName))}' -> '{resourceClaimUri}'");
```

axelmarquezh (changes requested):

axelmarquezh (approved):

PR #1005, opened 2024-03-26
[ODS-6310] Issue with routing around the identities hooks

axelmarquezh:


Application/EdFi.Ods.Api/Controllers/DataManagementControllerBase.cs
```diff
@@ -359,24 +359,6 @@ private EntityTagHeaderValue GetEtag(string etagValue)
             return new EntityTagHeaderValue(Quoted(etagValue));
         }
 
-        protected string GetResourceUrl()
```
axelmarquezh, line 362: Moved this method to `HttpRequestExtensions` so that we can call it from the `IdentitiesController`.

semalaiappan (approved):

PR #1011, opened 2024-03-29
[ODS-6296] XSD endpoint errors when extension feature is turned off

mjaksn:


Application/EdFi.Ods.Common/AssembliesProvider.cs
```diff
@@ -15,7 +16,7 @@ public class AssembliesProvider : IAssembliesProvider
     {
         private readonly ILog _logger = LogManager.GetLogger(typeof(AssembliesProvider));
 
-        public Assembly[] GetAssemblies() => AppDomain.CurrentDomain.GetAssemblies();
+        public Assembly[] GetAssemblies() => AssemblyLoadContext.Default.Assemblies.ToArray();
```
mjaksn, line 19: `AppDomain` is switched to `AssemblyLoadContext` because `AppDomain` has limited capabilities in .NET Core and does not have unloading capabilities (see first note on https://learn.microsoft.com/en-us/dotnet/api/system.appdomain?view=net-8.0). The collection returned by `AppDomain.CurrentDomain.GetAssemblies()`  includes assemblies loaded into a temporary `AssemblyLoadContext` instance to read metadata but were later unloaded, which is not the desired behavior here.

semalaiappan (approved):

PR #1016, opened 2024-04-04
[ODS-6321] Update Npgsql and HealthChecks libraries

vimayya:


JBrenesSimpat (commented):
Changes look good, ran locally initdev (MSSQL and Postgres), everything worked properly.

Please take a note of GitHub annotations about unchanged files:

- This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.
- The using directive for 'Dapper' appeared previously in this namespace

JBrenesSimpat (approved):

PR #1018, opened 2024-04-05
[ODS-3861] Parameter limit for API client associated with lots of nationwide districts/schools

axelmarquezh:
Note that I also modified `IntegrationTestHarness` to be able to specify a range of EdOrgs that a client should have access (see the [PR in the Implementation repo](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pull/980)). I use this new feature to assign 3k EdOrgs to the test client [here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/blob/19cb64ac7172fc704ff9db8d6796071ab9d4f681/logistics/scripts/modules/postmanTestHarnessConfiguration.json#L31).


Application/EdFi.Ods.Tests/EdFi.Ods.Security/Authorization/Repositories/EntityAuthorizerTests.cs
```diff
@@ -1,90 +0,0 @@
...
-using Test.Common;
-
-namespace EdFi.Ods.Tests.EdFi.Ods.Api.Security.Authorization.Repositories
-{
-    [TestFixture]
-    public class EntityAuthorizerTests
-    {
-        public class When_authorizing_an_unauthorized_caller: ScenarioFor<EntityAuthorizer>
-        {
-            private const long ExpectedClaimEndpointValue = long.MaxValue;
```
axelmarquezh, line 29: Removed this unit test since it's now being tested in Postman [here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/c4d330173271dc7d0a36874434653255c7064f1c/Postman%20Test%20Suite/Ed-Fi%20ODS-API%20Integration%20Test%20Suite.postman_collection.json#L12811).

Application/EdFi.Ods.Standard/Standard/4.0.0/Artifacts/MsSql/Structure/Ods/1006-AddSQLTypes-BigInt.sql
```diff
@@ -0,0 +1,9 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+CREATE TYPE dbo.BigIntTable AS TABLE
```
axelmarquezh, line 6: Note that I'm introducing BigIntTable for DS 4.

Although not ideal, it doesn't represent a change from the current behavior; for example the next query is generated when we call a `/deletes` endpoint (using the `main` branch), notice that the parameters are bigint. 
With my change, the parameter would be a TVP of `BigIntTable`.

Is this acceptable? If not, I can make the code changes to check the DS version and downcast to int32 if < 5.0.0.

image

Application/EdFi.Ods.Common/Infrastructure/Activities/ParameterListSetter.cs
```diff
@@ -23,5 +24,24 @@ public void SetParameterList(IQuery query, string name, IEnumerable ids)
...
+        /// <param name="propertyName">
+        /// The name of the Property in the class.
+        /// </param>
+        /// <param name="values">An array of values.</param>
+        /// <returns>An <see cref="AbstractCriterion" />.</returns>
+        public AbstractCriterion In(string propertyName, object[] values)
+        {
+            // Nhibernate's Restrictions.In() does not expect view aliases to be enclosed in
+            // curly brackets (contrary to Expression.Sql())
+            propertyName = propertyName
```
gmcelhanon, line 40: Since this code could be executed quite frequently, it could result in many unnecessary string allocations. Given that there is a fairly limited set of property names for which this would be repeatedly calculated, I think I would recommend save the result in a `ConcurrentDictionary<string, string>` using the `GetOrAdd` method with a factory method.

For the factory method, a few string allocations are probably not a big deal (using the code above), but you could also use a span here to perform 0 or 1 string allocation. The logic below code is definitely less readable, but I was already exploring it when I had the thought about just storing the result in a dictionary so I'm including it here anyway. So long as the string allocations only happen the first time a property name is passed in, I'm good with either implementation.

```csharp
var propertyNameSpan = propertyName.AsSpan();

int openBraceIndex = propertyNameSpan.IndexOf('{');
while (propertyNameSpan[openBraceIndex + 1] == '{') { openBraceIndex++; }

int closeBraceIndex = propertyNameSpan.LastIndexOf('}');
while (closeBraceIndex > 0 && propertyNameSpan[closeBraceIndex - 1] == '}') { closeBraceIndex--; }

if (closeBraceIndex >= 0)
{
    return propertyNameSpan.Slice(openBraceIndex + 1, closeBraceIndex - openBraceIndex - 1).ToString();
}
else if (openBraceIndex >= 0)
{
    return propertyNameSpan.Slice(openBraceIndex + 1).ToString();
}
else
{
    return propertyName;
}
```

Application/EdFi.Ods.Common/Infrastructure/SqlServer/SqlServerStructuredMappings.cs
```diff
@@ -0,0 +1,28 @@
...
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using NHibernate.Type;
+using System;
+using System.Collections.Generic;
+
+namespace EdFi.Ods.Common.Infrastructure.SqlServer
+{
+    public class SqlServerStructuredMappings
```
gmcelhanon, line 12: Probably makes sense to make this class `static` as well.

Application/EdFi.Ods.Standard/Standard/4.0.0/Artifacts/MsSql/Structure/Ods/1006-AddSQLTypes-BigInt.sql
gmcelhanon: I can't think of any big down sides to this approach (pun intended).

Application/EdFi.Ods.Common/Infrastructure/SqlServer/SqlServerTableValuedParameterListSetter.cs
```diff
@@ -58,24 +50,47 @@ public void SetParameterList(IQuery query, string name, IEnumerable ids)
...
+                             SqlServerStructuredMappings.StructuredTypeBySystemType[itemSystemType]);
+        }
+
+        /// <summary>
+        /// Creates a DataTabe with a single column "Id" populated with the given values.
+        /// </summary>
+        /// <param name="ids">The ids.</param>
+        /// <param name="idType">The type of the "Id" column.</param>
+        /// <returns></returns>
+        public static DataTable CreateIdDataTable(IEnumerable ids, Type idType)
```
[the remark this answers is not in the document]
gmcelhanon: I'd tend to agree. Maybe `SqlServerTableValuedParameterHelper`?

Application/EdFi.Ods.Common/Infrastructure/Activities/IParameterListSetter.cs
```diff
@@ -20,5 +21,16 @@ public interface IParameterListSetter
...
+        /// <summary>
+        /// Apply an "in" constraint to the named property.
+        /// </summary>
+        /// <param name="propertyName">
+        /// The name of the Property in the class. 
+        /// View aliases must be enclosed in curly brackets.
+        /// </param>
+        /// <param name="values">An array of values.</param>
+        /// <returns>An <see cref="AbstractCriterion" />.</returns>
+        public AbstractCriterion In(string propertyName, object[] values);
```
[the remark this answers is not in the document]
gmcelhanon: I would probably split the method into a new interface (`IMultiValueRestrictions` (?) to match the NHibernate class name (`Restrictions`) that we're essentially wrapping here) because there's no place where it's injected that both methods are used, right?

gmcelhanon (changes requested):

Application/EdFi.Ods.Common/Infrastructure/Activities/IParameterListSetter.cs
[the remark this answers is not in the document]
axelmarquezh: Right, there's no place where it's injected that both methods are used. I moved it to `IMultiValueRestrictions`.

Application/EdFi.Ods.Common/Infrastructure/SqlServer/SqlServerTableValuedParameterListSetter.cs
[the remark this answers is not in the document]
axelmarquezh: I moved it to `SqlServerTableValuedParameterHelper`.

Application/EdFi.Ods.Common/Infrastructure/Activities/ParameterListSetter.cs
axelmarquezh: I added a `ConcurrentDictionary` with a factory method that creates string allocations because it's more readable.

Application/EdFi.Ods.Common/Infrastructure/SqlServer/SqlServerStructuredMappings.cs
axelmarquezh: Fixed, thank you.

gmcelhanon (approved):

PR #1021, opened 2024-04-11
[ODS-6202] Switch to openapi-generator in SdkGen

ericstockdale:


JBrenesSimpat (approved):
All green on local test and run initdev

PR #1022, opened 2024-04-11
[ODS-6211] Visual Studio Template Support for VS 2022 and .NET6

ericstockdale:


ericstockdale:
The change to the relative path to License.txt does not feel right but it was the only way I could get it to work locally.

before
<License>..\..\..\LICENSE.txt</License>
after
<License>LICENSE.txt</License>

JBrenesSimpat (approved):

PR #1041, opened 2024-05-08
[ODS-6349] Profiles expiration doesn't clear mapping contracts

simpat-jesus:
MediatR registers all handler implementations as Transient, so I grouped all profile-related cached data objects to have the cache cleared by the same handler.

### DIFF
--- a/Application/EdFi.Ods.Features/Container/Modules/ProfilesModule.cs
+++ b/Application/EdFi.Ods.Features/Container/Modules/ProfilesModule.cs
@@ -79,10 +79,6 @@ namespace EdFi.Ods.Features.Container.Modules
                 .As<IAdminProfileNamesPublisher>()
                 .SingleInstance();
 
-            builder.RegisterType<ProfileMetadataCacheExpiredNotificationHandler>()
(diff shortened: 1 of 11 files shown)


Application/EdFi.Ods.Features/Container/Modules/ProfilesModule.cs
```diff
@@ -79,10 +79,6 @@ public override void ApplyConfigurationSpecificRegistrations(ContainerBuilder bu
                 .As<IAdminProfileNamesPublisher>()
                 .SingleInstance();
 
-            builder.RegisterType<ProfileMetadataCacheExpiredNotificationHandler>()
```
simpat-jesus, line 82: Handlers are registered [here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/7c8337a40b34c13825388f90545773a3b8c60acb/Application/EdFi.Ods.Api/Startup/OdsStartupBase.cs#L242). 
This change prevents registering a duplicated handler implementation.

mjaksn (approved):

PR #1043, opened 2024-05-09
[ODS-6352] Error 415 response should be updated to Problem Details RFC response

axelmarquezh:


Application/EdFi.Ods.Common/Exceptions/EdFiProblemDetailsExceptionBase.cs
```diff
@@ -76,9 +79,6 @@ namespace EdFi.Ods.Common.Exceptions;
   │    ├──┤ DatabaseConnectionException |
   │    │  └─────────────────────────────┘
   │    │  ┌──────────────────────────────┐
-  │    ├──┤ ClientConfigurationException |
```
axelmarquezh, line 79: We ended up not needing this Exception type; the type itself got removed in [#988](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/988) , but we missed removing this comment.

PR #1048, opened 2024-05-13
[ODS-6359] Rework on replacing usages of "resource" and "resource item" with "data" and "item" in the title and detail of Problem Details response

mjaksn:
This reverts changes made to contents of the errors string collection text in the first ODS-6359 PR. With this PR, only the Title and Detail portions of the error response object wording will use the terms "data" and "item" preferentially over  "resource" and "resource item". The contents of the Errors string collection portion of the API response, which generally contains details intended for more technical users, is reverted to use the original "resource" and "resource item" wording.


semalaiappan:
it was a couple of changes:
Back to old code/wording for just the contents of the string[] Errors property of the error json (details and title properties use new wording
Fix the changes so that it is a strict changing of the wording "resource" to "data" and "resource item" to "data item"
Vinaya wanted one exception to the strict wording change to make the wording in the authorization denied error sound better. "Access to the requested data could not be authorized." rather than "Access to the data could not be authorized."

semalaiappan (approved):

PR #1061, opened 2024-05-27
[ODS-6392] ODS/API v7.2 - issue when using id filter on swagger

axelmarquezh:


Postman Test Suite/Ed-Fi ODS-API ChangeQueries Key Changes and Deletes Test Suite.postman_collection.json
```diff
@@ -10571,7 +10571,7 @@
 											}
 										},
 										"url": {
-											"raw": "{{ApiBaseUrl}}/data/v3/ed-fi/staffSectionassociations/keyChanges?minChangeVersion={{known:{{scenarioId}}:initialChangeVersion}}",
+											"raw": "{{ApiBaseUrl}}/data/v3/ed-fi/staffSectionassociations/keyChanges?minChangeVersion={{known:{{scenarioId}}:newestChangeVersion}}",
```
axelmarquezh, line 10574: The `{{known:{{scenarioId}}:initialChangeVersion}}` variable doesn't exist, so it was sending an invalid filter all along.
The API continued processing the request, so the test seemed to be working as expected. After my change, the API detected the invalid parameter and started returning HTTP 400.
So, I updated the test to reflect what seemed to be the original intention.

Application/EdFi.Ods.Api/Startup/ApiBehaviorOptionsConfigurator.cs
```diff
@@ -34,7 +32,7 @@ public void Configure(ApiBehaviorOptions options)
     {
         options.InvalidModelStateResponseFactory = actionContext =>
         {
-            if (actionContext.ModelState.ValidationState == ModelValidationState.Invalid)
+            if (actionContext.ModelState.ErrorCount > 0)
```
axelmarquezh, line 35: The behavior started after [PR 903](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/903), which I _think_ is now modifying the `UrlQueryParametersRequest` after it's been validated. This changes the overall model state to `Unvalidated` (instead of `Invalid` if there was a validation error).
So, instead of checking the `ValidationState`, I now check if there are validation errors.

simpat-jesus (approved):

PR #1076, opened 2024-06-10
[ODS-6381] Updated Security Vulnerability Build for Released versions

semalaiappan:


mjaksn (approved):

vimayya (approved):
v7.1-patch1 is still needed. However since it will soon be switched to v7.1-patch2 with ODS-6390, I will merge this PR . v7.1-patch2 will be added with ODS-6390.

PR #1106, opened 2024-07-19
[ODS-6425] Support extensibility for read request authorization filtering based on custom database views

gmcelhanon:


Application/EdFi.Ods.Api/Container/Modules/ExtensionsModule.cs
```diff
@@ -64,27 +64,22 @@ public override void ApplyConfigurationSpecificRegistrations(ContainerBuilder bu
 
                     var relationshipContextDataProviderTypes = assembly.GetTypes()
                         .Where(
-                            t => !t.IsAbstract && typeof(IRelationshipsAuthorizationContextDataProvider<,>).IsAssignableFromGeneric(t))
+                            t => !t.IsAbstract && typeof(IRelationshipsAuthorizationContextDataProvider<>).IsAssignableFromGeneric(t))
```
gmcelhanon, line 67: Removed the unnecessary complexity of an old extensibility point (the generic type `TContextData` which was constrained to derive from `RelationshipsAuthorizationContextData`). This mechanism has never actually been used and was causing unnecessary complexity with the container registrations impacted by this ticket.

Application/EdFi.Ods.Api/Security/Authorization/ResourceClaimUriProvider.cs
```diff
@@ -54,15 +80,22 @@ public string[] GetResourceClaimUris(Type resourceType)
 
                     var schemaNameMap = schemaNameMapProvider.GetSchemaMapByProperCaseName(schemaProperCaseName);
 
-                    var uris = new[]
-                    {
-                        // Schema-based URI format
-                        CreateSchemaBasedResourceClaimUri(schemaNameMap.UriSegment, resourceName),
+                    // Include legacy representation for Ed-Fi Standard resources only
```
gmcelhanon, line 83: As written, this code was providing support for extension resources to use the URI format that does not include the schema segment. This runs counter to the guidance (and intent) regarding extension resources. Going forward, we will not be supporting matching extension resources to claim URIs without the schema appropriately included.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/AuthorizationStrategyNameAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+using EdFi.Ods.Common.Security.Authorization;
+
+namespace EdFi.Ods.Api.Security.AuthorizationStrategies;
+
+/// <summary>
+/// Provides the authorization strategy name that is associated with an <see cref="IAuthorizationStrategy" /> implementation
+/// (replacing the previously used class naming convention). 
+/// </summary>
+[AttributeUsage(AttributeTargets.Class, Inherited = false)]
+public sealed class AuthorizationStrategyNameAttribute : Attribute
```
gmcelhanon, line 16: We are no longer using a naming-convention based mapping from authorization strategy name (from security metadata) to the implementation class. This must now be specified explicitly through the use of this attribute. This change allows for the `CustomViewBasedAuthorizationStrategy` to exist with different behavioral characteristics from the original strategy implementations.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/CustomViewBased/CustomViewBasedAuthorizationStrategyProvider.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using Autofac.Extras.DynamicProxy;
+using EdFi.Ods.Common.Caching;
+using EdFi.Ods.Common.Models.Domain;
+using EdFi.Ods.Common.Security.Authorization;
+using log4net;
+
+namespace EdFi.Ods.Api.Security.AuthorizationStrategies.CustomViewBased;
+
+[Intercept(InterceptorCacheKeys.Security)]
```
gmcelhanon, line 10: This provider is cached with the security metadata, getting cleared when the security metadata is cleared.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/CustomViewBased/CustomViewBasisEntityProvider.cs
```diff
@@ -0,0 +1,58 @@
+using System.Collections.Concurrent;
+using System.Linq;
+using EdFi.Ods.Common.Conventions;
+using EdFi.Ods.Common.Models;
+using EdFi.Ods.Common.Models.Domain;
+
+namespace EdFi.Ods.Api.Security.AuthorizationStrategies.CustomViewBased;
+
+public class CustomViewBasisEntityProvider : ICustomViewBasisEntityProvider
```
gmcelhanon, line 9: To avoid the need for security metadata database changes, we use an authorization strategy naming convention for custom view-based authorization strategies. It uses the following format: `{entityName}With...`. The entity name is used to find a matching entity in the Ed-Fi Standard and failing that it looks in extension schemas (in no particular order) until it finds a matching entity. The primary key of that entity then is deemed to be the basis for joining with the authorization subject (the focus of the current requested resource).

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/NamespaceBased/NamespaceBasedAuthorizationFilterDefinitionsFactory.cs
```diff
@@ -61,10 +68,18 @@ public IReadOnlyList<AuthorizationFilterDefinition> CreateAuthorizationFilterDef
     private static void ApplyAuthorizationCriteria(
         ICriteria criteria,
         Junction @where,
-        string subjectEndpointName,
+        string[] subjectEndpointNames,
```
gmcelhanon, line 71: Whereas joins for authorization have until now been on a single column, the custom view-based authorization opens us up to the possibility of composite key joins with multiple columns. Thus, the technical artifacts have been modified as needed to support this.

Application/EdFi.Ods.Api/Security/AuthorizationStrategies/Relationships/Filters/ICriteriaExtensions.cs
```diff
@@ -92,5 +96,76 @@ private static string GetFullNameForView(this string viewName)
...
+            if (subjectEndpointNames.Length == 1)
+            {
+                // Apply authorization join using ICriteria
+                criteria.CreateEntityAlias(
+                    authViewFullAlias,
+                    Restrictions.EqProperty(
+                        $"aggregateRoot.{subjectEndpointNames[0]}",
+                        $"{authViewFullAlias}.{viewTargetEndpointNames[0]}"),
+                    joinType,
+                    customViewBasedAuthorizationStrategy.BasisEntity.EntityTypeFullName());
```
gmcelhanon, line 136: NHibernate only supports joins to _mapped_ entities (that also don't have any mapped associations) using `ICriteria`. However, `ICriteria`, as with any other NHibernate-based query mechanism can only work with mapped entities, but given the dynamic nature of the custom views (which can be added after code generation to created corresponding entity classes and ORM mappings AND can even be added at runtime and refreshed with security metadata), we have to use some "creative" processing. For now, the query is generated to join with the "basis" entity identified based on the custom authorization view name. The generated SQL will be manipulated just prior to execution to replace the "basis" entity with the custom authorization view. See the updated NHibernate interceptor implementation.

Application/EdFi.Ods.Common/Caching/BytesExtensions.cs
```diff
@@ -109,7 +115,7 @@ private static byte[] GetArrayBytes(List<byte[]> byteArrays)
             foreach (byte[] valueAsBytes in byteArrays)
             {
                 valueAsBytes.AsSpan().CopyTo(buffer.AsSpan(startIndex));
-                _enumerableItemDelimiter.AsSpan().CopyTo(buffer.AsSpan(valueAsBytes.Length + 1));
+                _enumerableItemDelimiter.AsSpan().CopyTo(buffer.AsSpan(startIndex + valueAsBytes.Length));
```
gmcelhanon, line 118: @vimayya This is a bugfix discovered during development of this feature.

Application/EdFi.Ods.Common/Infrastructure/Filtering/AuthorizationFilterDefinition.cs
```diff
@@ -43,7 +42,7 @@ public class AuthorizationFilterDefinition
         public AuthorizationFilterDefinition(
             string filterName,
             string friendlyHqlConditionFormat,
-            Action<ICriteria, Junction, string, IDictionary<string, object>, JoinType> criteriaApplicator,
+            Action<ICriteria, Junction, string[], IDictionary<string, object>, JoinType, IAuthorizationStrategy> criteriaApplicator,
```
gmcelhanon, line 45: `IAuthorizationStrategy` was added here because with the custom view-based authorization, we create them using a factory (so there will be multiple instances of the class, rather than a singleton) and each strategy instance carries additional contextual information relevant to performing authorization using the custom view.

Application/EdFi.Ods.Common/Infrastructure/Interceptors/EdFiOdsInterceptor.cs
```diff
@@ -0,0 +1,69 @@
...
+                return isTransient;
+            }
+
+            return base.IsTransient(entity);
+        }
+
+        public override SqlString OnPrepareStatement(SqlString sql)
+        {
+            // Determine if we need to perform custom authorization view SQL manipulation
+            if (sql.IndexOf(CustomViewHelpers.CustomViewAliasPrefixBase, 0, sql.Length, StringComparison.Ordinal) >= 0)
```
gmcelhanon, line 41: Replaces the basis entity with the custom authorization view in the generated SQL, just prior to execution.

Application/EdFi.Ods.Common/Infrastructure/Interceptors/EdFiOdsInterceptor.cs
```diff
@@ -0,0 +1,69 @@
...
+            RegexOptions.Compiled);
+
+        public override bool? IsTransient(object entity)
+        {
+            var property = entity.GetType()
+                                 .GetProperty("CreateDate");
+
+            if (property != null)
+            {
+                bool isTransient = Convert.ToDateTime(property.GetValue(entity)) == default(DateTime);
```
mjaksn, line 30: I think the exception-handling infrastructure may need to be updated slightly to process an exception thrown from `property.GetValue()` (or possibly any exception thrown below the NHibernate interceptor invocation). From a quick check, it appears that any exception thrown in this context will be wrapped in a `GenericADOException` by NHibernate, which would prevent an Ed-Fi custom exception type from being correctly serialized into an API response. Also, having the outer exception type of `GenericADOException` showing first in logs/stack traces for cases where the exception is caused by a naming mismatch in the auth view configuration might make troubleshooting less straightforward.

mjaksn (changes requested):

Application/EdFi.Ods.Common/Security/CustomViewBased/CustomViewHelpers.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// <returns>The view name corresponding to the supplied alias prefix.</returns>
+    /// <exception cref="Exception">Occurs when the view name cannot be identified for the supplied alias prefix.</exception>
+    public static string GetViewName(string aliasPrefix)
+    {
+        if (_viewNameByAuthViewAliasPrefix.TryGetValue(aliasPrefix, out string viewName))
+        {
+            return viewName;
+        }
+
+        throw new Exception($"Unable to find view name for alias prefix '{aliasPrefix}'.");
```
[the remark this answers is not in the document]
gmcelhanon: No... this utility function is providing the view name that is associated with the custom view authorization alias that is encountered by the interceptor during final SQL preparation. The alias that was encountered was also previously generated by the other utility function on this class and the association between the custom view name and the alias has been cached in a dictionary.

So, during final SQL preparation, the only aliases encountered will be the ones that were previously generated and they should always be found.

In essence, this exception should never happen, and it does not reflect a security configuration issue. An internal server error is more appropriate if for some reason a match failed.

Application/EdFi.Ods.Standard/Standard/5.1.0/Security/Authorization/Overrides/ProgramEvaluationObjectiveRelationshipsAuthorizationContextDataProvider.cs
```diff
@@ -15,20 +15,19 @@ namespace EdFi.Ods.Standard.Security.Authorization.Overrides
     /// Creates and returns an <see cref="RelationshipsAuthorizationContextData"/> instance for making authorization decisions for access to the edfi.ProgramEvaluationObjective table of the ProgramEvaluationObjective aggregate in the Ods Database.
     /// </summary>
     [ExcludeFromCodeCoverage]
-    public class ProgramEvaluationObjectiveRelationshipsAuthorizationContextDataProvider<TContextData> : IRelationshipsAuthorizationContextDataProvider<IProgramEvaluationObjective, TContextData>
-        where TContextData : RelationshipsAuthorizationContextData, new()
+    public class ProgramEvaluationObjectiveRelationshipsAuthorizationContextDataProvider : IRelationshipsAuthorizationContextDataProvider<IProgramEvaluationObjective>
     {
         /// <summary>
         /// Creates and returns an <see cref="TContextData"/> instance based on the supplied resource.
```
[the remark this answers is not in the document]
gmcelhanon: Clean this usage up.

Application/EdFi.Ods.Api/Security/Authorization/AuthorizationBasisMetadataSelector.cs
```diff
@@ -171,13 +131,17 @@ IReadOnlyList<IAuthorizationStrategy> GetAuthorizationStrategies(IReadOnlyList<s
...
+                            var authorizationStrategy = authorizationStrategyProvider.GetByName(strategyName);
+
+                            if (authorizationStrategy != null)
+                            {
+                                return authorizationStrategy;
+                            }
                         }
 
-                        return _authorizationStrategyByName[strategyName];
+                        throw new Exception($"Could not find an authorization implementation for authorization strategy name '{strategyName}'.");
```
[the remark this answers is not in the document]
gmcelhanon: Yes, I think this one should be a `SecurityConfigurationException` and I found another one above it in the same function. I'm converting both of them from `Exception`.

mjaksn (approved):

PR #1108, opened 2024-07-19
[ODS-6440] Sandbox Admin's "Change Secret" button fails after EFCore upgrade - v5

axelmarquezh:


Application/EdFi.Admin.DataAccess/Repositories/ClientAppRepo.cs
```diff
@@ -199,6 +199,16 @@ public async Task<ApiClient> GetClientAsync(string key)
             }
         }
 
+        public ApiClient GetClientIncludingApplication(string key)
```
axelmarquezh, line 202: This new method only includes the necessary fields for the `Change Secret` button.
More specifically, the issue gets fixed by not including the `ApplicationEducationOrganizations`.

semalaiappan (approved):

PR #1113, opened 2024-07-25
[ODS-6407] Improve workflow execution

axelmarquezh:


.github/workflows/Trgr InitDev workflows in Implementation repo.yml
```diff
@@ -3,7 +3,6 @@ name: Trgr InitDev workflows in Implementation repo
 on:
   pull_request:
     branches: [main, 'b-v*-patch*','feature-*']
-  workflow_dispatch:
```
axelmarquezh, line 6: Is there a use case that would benefit from allowing this workflow to be manually executed?

.github/workflows/Lib edFi.ods.standard.yml
```diff
@@ -34,7 +34,6 @@ jobs:
     with:
       calling_branch: ${{ github.head_ref || github.ref_name }}
   build:
-    if: ${{ always() }}
```
axelmarquezh, line 37: Is there a reason to use `if: ${{ always() }}`?
I'm removing it since it [makes the workflows un-cancellable](https://docs.github.com/en/actions/learn-github-actions/expressions#always) and interferes with the `concurrency` setting.

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
semalaiappan: No

stephenfuqua: Personally, I usually like to leave `workflow_dispatch` for manual re-execution or experiment on an alternate branch.

simpat-jesus (approved):
Using a label to run the PR checks is very clever, good one!

.github/workflows/Trgr InitDev workflows in Implementation repo.yml
axelmarquezh: I added `workflow_dispatch` back to allow for manual re-executions/experimentation.

semalaiappan (approved):

PR #1149, opened 2024-09-25
[ODS-4799] Introduce resource POST/Retry order in dependency endpoint

gmcelhanon:


Application/EdFi.Ods.Api/Models/GraphML/GraphMLEdge.cs
```diff
@@ -3,19 +3,71 @@
...
 
+using System;
+using EdFi.Ods.Common.Models.Domain;
+using QuickGraph;
+
 namespace EdFi.Ods.Api.Models.GraphML
 {
     // ReSharper disable once InconsistentNaming
-    public class GraphMLEdge
+    public class GraphMLEdge : IEquatable<GraphMLEdge>, IEdge<GraphMLNode>
```
gmcelhanon, line 13: The GraphML model used for serialization needed to be processed as a bidirectional graph with the `BreakCycles` logic (an extension method written for bidirectional graphs) after introducing the `/ed-fi/staffs#Retry` node into the graph.

In moving the dependencies of the staff employment/assignment associations to be dependents instead of the new "Retry' node, a "soft" (optional) circular dependency on between the Staff "Retry" node and the StaffEdOrgAssignmentAssociation resulted because of the StaffEdOrgAssignmentAssociation's optional dependency on the StaffEdOrgEmploymentAssociation. This circular dependency needed to be removed reusing the `BreakCycles` logic mentioned above, and requiring GraphML models to support the behavior required by the graph library (`IEquatable<T>` and for the node, `IComparable<T>`).

image

mjaksn (approved):

PR #1155, opened 2024-10-08
[ODS-6520] Bring in first set of DS 5.2 updates

semalaiappan:


mjaksn:
@semalaiappan , a folder with DS 5.2 approval test files needs to be added under ./Ed-Fi-ODS/tests/EdFi.Ods.WebApi.IntegrationTests/Controllers/Standard/

PR #1156, opened 2024-10-10
[ODS-5665] FOR PRELIMINARY REVIEW

gmcelhanon:


Application/EdFi.Ods.Common/Providers/Queries/Criteria/IdentificationCodeAggregateQueryCriteriaApplicator.cs
```diff
@@ -0,0 +1,134 @@
...
+        DatabaseEngine databaseEngine)
+    {
+        _descriptorResolver = descriptorResolver;
+        _databaseEngine = databaseEngine;
+    }
+
+    public void ApplyAdditionalParameters(QueryBuilder queryBuilder, Entity entity, AggregateRootWithCompositeKey specification,
+        IDictionary<string, string> additionalParameters)
+    {
+        if (additionalParameters == null || !additionalParameters.Any() || additionalParameters.All(
```
gmcelhanon, line 37: I think you could leave the `null` and the `Any` checks, but I don't think the latter is needed because ASP.NET will fill that dictionary with values that were not bindable to the other models in the controller signature. We should confirm this, but I think this will be wasted worked. Thoughts/concerns?

Application/EdFi.Ods.Common/Providers/Queries/AggregateRootCriteriaProviderHelpers.cs
```diff
@@ -26,4 +30,36 @@ public static string[] GetUniqueIdProperties(IPersonTypesProvider personTypesPro
...
+                _entityIdentificationCodePropertyCache[entityFullName] = identificationCodeEntityProperty;
+                return identificationCodeEntityProperty;
+            }
+
+            entity = entity.BaseEntity;
+        }
+
+        return null;
+
+        EntityProperty FindIdentificationCodePropertyInAssociations(Entity entity)
```
gmcelhanon, line 57: This method name is misleading because you're looking for a property in the child table, not the association (aka FK). The identification code will never be part of any association. So if we were leaving this as-is (but I recommend changes to what this is returning below), I would say call this "TryFindIdentificationCodeProperty" and change the semantics (return a `bool` for success/failure, and an `out` property). But... read on.

Application/EdFi.Ods.Common/Providers/Queries/Criteria/IdentificationCodeAggregateQueryCriteriaApplicator.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        string GetParameterNameForIdentificationCodeProperty(EntityProperty property)
+        {
+            return property.DescriptorEntity?.Name ?? property.PropertyName;
+        }
+        
+        bool IsQueryableIdentificationCodeProperty(EntityProperty property)
+        {
+            return property.PropertyName.Equals("IdentificationCode") ||
+                   !(property.IsFromParent || property.IsAlreadyDefinedInCSharpEntityBaseClasses());
```
gmcelhanon, line 131: `IsAlreadyDefinedInCSharpEntityBaseClasses` is currently unused by the API project and probably should only be available for use by CodeGen. It looks like `IsPredefinedProperty` is in use and accomplishes the same thing. Let's use that.

Application/EdFi.Ods.Common/Providers/Queries/Criteria/IdentificationCodeAggregateQueryCriteriaApplicator.cs
```diff
@@ -0,0 +1,134 @@
...
+        IEnumerable<string> GetIdentificationCodeEntityTableJoinColumnNames(Entity identificationCodeEntity,
+            DatabaseEngine databaseEngine)
+        {
+            return identificationCodeEntity.Identifier.Properties
+                .Where(p => p.IsFromParent && p.IsIdentifying)
+                .Select(p => p.ColumnName(databaseEngine, p.PropertyName));
+        }
+    }
+
+    private IDictionary<string, EntityProperty> GetIdentificationCodeParameters(EntityProperty entityIdentificationCodeProperty)
```
gmcelhanon, line 117: Do this work once per entity, and cache it by the root entity name in the top-level `ConcurrentDictionary`.

PR #1157, opened 2024-10-10
[ODS-6502] Update security metadata for StudentContactAssociation

semalaiappan:


axelmarquezh (changes requested):
Please also apply these changes to DS 4x.

axelmarquezh (approved):

PR #1205, opened 2025-01-13
[ODS-6602] API only allows tenant1's profiles when different profiles are set up for different tenants

mjaksn:
The Admin database changes associated with this ticket are in a separate PR - #1212.


Application/EdFi.Ods.Profiles.Test/ProfilesTenant2.xml
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+
+<Profiles>
+  <!-- The Tenant1 version of this profile does not include the 'SchoolTypeDescriptor' property in its ReadContentType, and this version for Tenant2 does -->
+  <!-- The Tenant1 version of this profile includes the 'AdministrativeFundingControlDescriptor' property in its ReadContentType, and this version for Tenant2 does not -->
+  <!-- Resource-level IncludeOnly/ExcludeOnly profiles -->
+  <Profile name="Tenant-Two-Test-Profile">
```
axelmarquezh, line 7: Please also add Postman tests for the scenario where both tenant1 and tenant2 have a profile that share the same name (but different definitions).

As of now, that scenario doesn't work because we are hitting a different cache ([here](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/blob/a18ba8e0132360f7db2a6396779b6e1337a4ae27/Application/EdFi.Ods.Api/Serialization/ProfilesAwareContractResolver.cs#L88)) that isn't tenant aware (so it caches the first profile that gets requested).

Artifacts/MsSql/Structure/Admin/0177-Add-Unique-Constraint-Profiles.sql
```diff
@@ -0,0 +1,12 @@
+-- SPDX-License-Identifier: Apache-2.0
+-- Licensed to the Ed-Fi Alliance under one or more agreements.
+-- The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+-- See the LICENSE and NOTICES files in the project root for more information.
+
+ALTER TABLE dbo.Profiles
+ALTER COLUMN ProfileName NVARCHAR(500) NOT NULL
```
axelmarquezh, line 7: Should we also apply 500 limit to PGSQL to keep both DB engines consistent?
CC: @vimayya

axelmarquezh (changes requested):

Application/EdFi.Ods.Common/Metadata/Profiles/ProfileMetadataProvider.cs
```diff
@@ -29,8 +29,27 @@ public ProfileMetadataProvider(IProfileDefinitionsProvider[] profileDefinitionsP
...
+
+            var profilesWithDuplicateDefinitions = profileDefinitions
+                .GroupBy(x => x.Key, StringComparer.OrdinalIgnoreCase)
+                .Where(x => x.Count() > 1)
+                .SelectMany(x => profileDefinitions.Where(y => y.Key.Equals(x.Key, StringComparison.OrdinalIgnoreCase)))
+                .ToArray();
+
+            if (profilesWithDuplicateDefinitions.Any())
+            {
+                var duplicateProfileNames = string.Join(", ", profilesWithDuplicateDefinitions.Select(x => x.Key).Distinct());
```
axelmarquezh, line 44: `.Distinct()` is case-sensitive while the checks we do a couple of lines above are not, which can potentially output messages like:
`The following profile names were not loaded because multiple XML definitions were provided: Test-Profile, test-profile`

Application/EdFi.Ods.Common/Metadata/Profiles/ProfileMetadataProvider.cs
```diff
@@ -29,8 +29,27 @@ public ProfileMetadataProvider(IProfileDefinitionsProvider[] profileDefinitionsP
...
+
+                _logger.Error(
+                    $"The following profile names were not loaded because multiple XML definitions were provided: {duplicateProfileNames}");
```
axelmarquezh, line 47: This validation detects and logs duplicate profiles (within a tenant) as expected. However, if I try to retrieve a resource that has such profile enforced, it fails with HTTP 500 (because it can't find the profile).

What should be the expected behavior?
CC: @vimayya

Application/EdFi.Ods.Profiles.Test/EdFi.Ods.Profiles.Test.csproj
```diff
@@ -16,8 +16,14 @@
...
+    <None Remove="ProfilesTenant2.xml" />
+    <EmbeddedResource Include="ProfilesTenant2.xml" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\..\Ed-Fi-ODS\Application\EdFi.Ods.Common\EdFi.Ods.Common.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="ProfilesTenant1.xml" />
+    <EmbeddedResource Include="ProfilesTenant1.xml" />
+  </ItemGroup>
```
axelmarquezh, line 28: Any reason to add `ProfilesTenant1.xml` and `ProfilesTenant2.xml` as embedded profiles? They're manually being loaded in Test Harness.

axelmarquezh (changes requested):

Postman Test Suite/Multitenancy/Ed-Fi ODS-API Multi-Tenancy.postman_collection.json
```diff
@@ -1690,6 +1690,590 @@
...
+											"});\r",
+											"\r",
+											"pm.test(\"Response should not contain a property not included in the Tenant2 test profile\", () => {\r",
+											"    var responseData = pm.response.json();\r",
+											"    pm.expect(responseData).to.not.have.property(pm.variables.replaceIn(\"{{supplied:tenant2:nonIncludedPropertyName}}\"));\r",
+											"});\r",
+											"\r",
+											"pm.test(\"The nameOfInstitution property of the response should match the expected value for Tenant1\", () => {\r",
+											"    var responseData = pm.response.json();\r",
+											"    pm.expect(responseData).to.have.property(pm.variables.replaceIn(\"{{supplied:tenant2:includedPropertyName}}\"));\r",
```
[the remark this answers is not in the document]
mjaksn: I removed the first check since it was redundant.

Application/EdFi.Ods.Profiles.Test/ProfilesTenant2.xml
mjaksn: I added same profile name (but different definitions) postman tests and updated the MapptingContract caching to be tenant-aware.

Artifacts/MsSql/Structure/Admin/0177-Add-Unique-Constraint-Profiles.sql
mjaksn: I added the corresponding column type change to the PGSQL script.

Application/EdFi.Ods.Common/Metadata/Profiles/ProfileMetadataProvider.cs
mjaksn: This now uses `StringComparison.OrdinalIgnoreCase` for the comparison.

Application/EdFi.Ods.Common/Metadata/Profiles/ProfileMetadataProvider.cs
mjaksn: The 500 error is the same error that is currently generated if a client is assigned a profile with an invalid XML definition in the Admin database (which is different than the 406 error if the profile is specified in the content-type header). @vimayya , should we return a 406 in both cases? The reason for the difference is that the enforcement of assigned profiles happens in a middleware separate from the logic for looking up a profile specified in the content-type header.

vimayya: Profile XML misconfiguration and duplicates in profile names are issues that can't be fixed by API clients. So I think 500 level error makes sense.

axelmarquezh (changes requested):

axelmarquezh (approved):

PR #1237, opened 2025-07-09
[ODS-6653] Add Data Standard v6.0 to build matrix

semalaiappan:


axelmarquezh:
Will be handled in #1240.

PR #1277, opened 2025-10-03
[ODS-6701] surveyDomain should not have a parent resource claim 

semalaiappan:


simpat-jesus (changes requested):
Can we rename 2600 scripts name? Having Standard 6 scripts in 4.0.0/5.2.0 folder can be confusing

PR #1323, opened 2026-03-27
[ODS-6819] Security updates for TokenInfoController

stephenfuqua:
## Summary

- **Finding 1 (Medium):** Collapse `404`/`401` into a single `404` for tokens that don't exist or belong to another client, eliminating an enumeration oracle that could expose valid access token GUIDs owned by other vendors.
- **Finding 2 (Low–Medium):** In JWT mode, validate that `tokenInfoRequest.Token` matches the raw Bearer token from the `Authorization` header before proceeding. Previously any non-null string was accepted and silently ignored, violating RFC 7662 §2 semantics.
- **Finding 3 (Low):** Normalize all `BadRequest` error message detail strings to identical text so the backend token mode (GUID vs JWT) is not leaked to callers.
- Adds `docs/security-token-info-controller.md` documenting all findings and their dispositions.

## Test plan

- [ ] GUID mode: POST `/oauth/token_info` with a valid token owned by the authenticated client → `200 OK`
- [ ] GUID mode: POST with a GUID that does not exist → `404 Not Found`
- [ ] GUID mode: POST with a GUID belonging to a different client → `404 Not Found` (was `401` before this fix)
- [ ] GUID mode: POST with a non-GUID string → `400 Bad Request` with generic message (no mention of "GUID")
- [ ] JWT mode: POST with the correct Bearer JWT as the `token` body param → `200 OK`
- [ ] JWT mode: POST with a different/garbage `token` value → `400 Bad Request`
- [ ] JWT mode: POST with null `token` → `400 Bad Request` with the same message text as GUID mode failure
- [ ] Feature disabled: any POST → `404 Not Found`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Utilities/CodeGeneration/EdFi.Ods.CodeGen.Tests/IntegrationTests/Providers/AssemblyDataProviderTests.cs
stephenfuqua, line 1: How did this error get into `main`?

I was unable to build the code until I made the changes below.

Application/EdFi.Ods.Features/Controllers/TokenInfoController.cs
stephenfuqua, line 1: This update is, naturally, causing a Postman test failure. Currently getting a 404 with an invalid token. Need to look at this more closely: should it be a 400 (what it was before), or a 401? @stephenfuqua to analyze before taking this out of draft. 404 is definitely _not_ an appropriate response for a POST to an endpoint that _does_ exist.

Image

Application/EdFi.Ods.Features/Controllers/TokenInfoController.cs
```diff
@@ -84,42 +84,62 @@ private async Task<IActionResult> GetTokenInformation(TokenInfoRequest tokenInfo
...
+                if (!Guid.TryParse(_httpContextAccessor.HttpContext?.User.FindFirst(JwtRegisteredClaimNames.Jti)?.Value, out guidAccessToken))
+                {
+                    return BadRequest(
+                        new BadRequestException(
+                            "An invalid token was provided",
+                            new[] { "The token was not present, or was not processable." })
+                        {
+                            CorrelationId = _logContextAccessor.GetCorrelationId()
+                        }.AsSerializableModel());
+                }
```
stephenfuqua, line 114: Two identical failure blocks above - the `if` statements can be recombined.

Application/EdFi.Ods.Common/Extensions/LogContextAccessorExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static class LogContextAccessorExtensions
     /// <returns>The previously set correlation id.</returns>
     public static string GetCorrelationId(this ILogContextAccessor logContextAccessor)
     {
-        return (string)logContextAccessor.GetValue(CorrelationConstants.LogContextKey);
+        return logContextAccessor.GetValue(CorrelationConstants.LogContextKey) as string;
```
stephenfuqua, line 20: This seemed rather strange to me. I had to ask Claude to explain, since I had never heard of these two forms being different. Claude's response is quite interesting.

---

Both attempt to convert an object to string, but they behave differently when the value is not a string:

- (string)obj — hard cast. If the value is null or any non-string type, it throws an InvalidCastException at runtime.
- obj as string — safe cast. If the value is null or a non-string type, it returns null instead of throwing.

In this context, GetValue returns object, and the key may not be set yet (e.g., early in the request lifecycle before the correlation ID has been assigned). In that case the value would be null. The as cast handles that gracefully by returning null, while the hard cast would throw.

The as form is also idiomatic for "give me this as a string if you can, otherwise null" — which matches what callers of GetCorrelationId() would expect: either a string or nothing.

Application/EdFi.Ods.Features/Controllers/TokenInfoController.cs
```diff
@@ -84,42 +84,54 @@ private async Task<IActionResult> GetTokenInformation(TokenInfoRequest tokenInfo
                 return BadRequest(
                     new BadRequestException(
                         "An invalid token was provided",
-                        new[] { "The token was not present, or was not processable as a GUID value." })
+                        new[] { "The token was not present, or was not processable." })
```
stephenfuqua, line 87: We don't need to tell attackers that token's are GUIDs.

Application/EdFi.Ods.Features/Controllers/TokenInfoController.cs
```diff
@@ -84,42 +84,54 @@ private async Task<IActionResult> GetTokenInformation(TokenInfoRequest tokenInfo
...
+                // Validate that the submitted token matches the Bearer token used to authenticate
+                // this request (RFC 7662 §2 — the caller must submit the token to be introspected).
+                var authHeader = _httpContextAccessor.HttpContext?.Request.Headers[HeaderNames.Authorization].ToString();
+                var bearerToken = authHeader?.StartsWith("Bearer ", StringComparison.OrdinalIgnoreCase) == true
+                    ? authHeader["Bearer ".Length..].Trim()
+                    : null;
+
+                if (bearerToken == null || 
+                    !string.Equals(tokenInfoRequest.Token, bearerToken, StringComparison.Ordinal) ||
+                    !Guid.TryParse(_httpContextAccessor.HttpContext?.User.FindFirst(JwtRegisteredClaimNames.Jti)?.Value, out guidAccessToken)
```
stephenfuqua, line 104: safer handling of the token

Application/EdFi.Ods.Features/Controllers/TokenInfoController.cs
```diff
@@ -84,42 +84,54 @@ private async Task<IActionResult> GetTokenInformation(TokenInfoRequest tokenInfo
...
+                        }.AsSerializableModel());
+                }
             }
 
             var oAuthTokenClientDetails = await _apiClientDetailsProvider.GetApiClientDetailsForTokenAsync(guidAccessToken.ToString("N"));
 
-            if (oAuthTokenClientDetails == null)
-            {
-                return NotFound();
-            }
```
stephenfuqua, line 110: It is true that the token was not found, but since the _endpoint_ exists, 404 is not an appropriate response. 401 is the appropriate response.

Application/EdFi.Ods.Tests/EdFi.Ods.Features/Controllers/TokenInfoControllerTests.cs
stephenfuqua, line 1: There were no unit tests before.

Postman Test Suite/Ed-Fi ODS-API Integration Test Suite ResponseTests.postman_collection.json
stephenfuqua, line 1: The old file had tabs in it, but Postman is now exporting with spaces. That is why the diff is too large to show in GitHub.

Actual changes are below, from Winmerge. The _updated_ version is accidentally on the left side. 

In this first one, the change to expected text is cutoff. Just shows removing the reference the token needing to be a Guid. I changed the name of the test because it was misleading: it sends an empty JSON payload - not just an invalid token, but the value isn't even there.

Image

Now here's a real request with a valid-looking (Guid) token that is not actually valid.

Image

stephenfuqua:
Claude flagged this as a security risk due to enumeration attacks. I’m not sure that I really reflected deeply enough on the actual use case though. It suddenly occurs to me that- and should have before - that 401 is not appropriate when the bearer token is correct but the payload contains an invalid token. Indeed, 400 seems more appropriate after all. But I need to go read the introspection specification again before drawing any permanent conclusion. At this moment I’m not sure why the payload even has a token different than the auth header. 

[Token introspection specification](https://datatracker.ietf.org/doc/html/rfc7662) points out that Token Introspection is typically for server to server use, where server A uses its own token to ask Server B if a client’s token is still valid. We are not using it this way - we are instead using it for client’s to lookup additional information about their own token. 

In this sense, perhaps it would be better to either break with the spec and just ignore the payload’s token. Indeed, I think @vimayya mentioned that very idea in a recent conversation, though I had not connected the dots yet. That seems like a better solution than the one I have taken. 

In the DMS the token introspection potentially applies, if DMS needs to ask the Configuration Service about validity of the incoming client token. But this should not be necessary since the client token is a signed JWT; DMS should be able to validate that signature. Introspection only valuable if we want to check that a given token has not been invalidated after it was issued. I don’t recall if we have opted for that approach. If we did go that direction, then per the following statement from the introspection specification, the response code for an invalid token in the payload should be 200 with active = false. 

> Note that a properly formed and authorized query for an inactive or otherwise invalid token (or a token the protected resource is not allowed to know about) is not considered an error response by this specification.  In these cases, the authorization server MUST instead respond with an introspection response with the "active" field set to "false" as described in [Section 2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).

Issue #1343, opened 2026-05-30
Redis External Cache: Timeouts and Poor Performance Due to Implementation Issues


stephenfuqua:
## Executive Summary

The Ed-Fi ODS/API Redis caching implementation contains several issues that collectively may explain reported timeouts and poor performance when Redis is enabled. The most critical problems are: **synchronous blocking calls for the descriptor cache path**, **lack of Redis connection tuning**, **fire-and-forget expiration commands that silently fail**, and **an N+1-style problem with per-request cache hits at high volume**. Additionally, there is an opportunity to provide a hybrid cache scenario with L1 (in-memory) falling back to L2 (Redis) before gracefully degrading to L3 (RDBMS).

### 1. Synchronous IDistributedCache Calls in Descriptor Cache Path (HIGH SEVERITY)

**Location:** `ExternalCacheProvider.cs` lines 54, 103, 120

The `ExternalCacheProvider<TKey>` uses the **synchronous** `IDistributedCache` methods:

- `_distributedCache.GetString(keyAsString)` (line 54)
- `_distributedCache.SetString(...)` (lines 103, 120)

These block the calling thread while waiting for a Redis round-trip. The descriptor cache is invoked via the `CachingInterceptor` (a Castle DynamicProxy `IInterceptor`), which has a synchronous `Intercept()` method — there is no async path available.

**Impact:** As the documentation explains, a single paged request for 500 `studentSchoolAssociations` can generate **3000+ descriptor cache hits**. Each of those hits becomes a synchronous, blocking Redis round-trip. Even at 0.5ms per call, that's **1.5 seconds of latency** added to a single request, purely from descriptor resolution. Thread pool starvation under concurrent load makes this exponentially worse.

### 2. No Redis Connection Configuration Tuning (HIGH SEVERITY)

**Location:** `RedisConnectionProvider.cs`, `OverrideRedisExternalCacheModule.cs`

The `RedisConnectionProvider` connects using the raw configuration string (e.g., `"localhost"`) with no customization of:

- `syncTimeout` (default 5000ms — easy to hit under load)
- `asyncTimeout` (default 5000ms)
- `connectTimeout` (default 5000ms)
- `connectRetry` (default 3)
- `abortOnConnectFail` (default true in StackExchange.Redis — connection failures are fatal)
- `keepAlive` (default 60s)

[...]

PR #1344, opened 2026-06-07
[ODS-6847] Redis Refactors: Improved Configuration, Resilience, Async Usage

stephenfuqua:
This PR attempts to reduce Redis timeout and performance problems by:

- Introducing a Level 1 (in memory) / Level 2 (external) caching paradigm
- Changing the Descriptor cache to run with full `async` support
- Exposing more cache configuration settings for system tuning
- Adding Polly-based resilience (retry, backoff, circuit breaker)

There are a few opportunistic refactorings for reducing code duplication and for improving use of modern C# syntax and formatting; however, this was not a key goal of the PR and it does not attempt to fix / reformat every touched file.

Addresses #1343 

Pairs with [Ed-Fi-ODS-Implementation PR 1387](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pull/1387)

### DIFF
--- a/Application/EdFi.Ods.Api/Caching/Person/IDistributedLockProvider.cs
+++ b/Application/EdFi.Ods.Api/Caching/Person/IDistributedLockProvider.cs
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace EdFi.Ods.Api.Caching.Person;
+
+/// <summary>
+/// Defines methods for coordinating distributed access to background cache initialization.
+/// </summary>
+public interface IDistributedLockProvider
+{
+    /// <summary>
+    /// Attempts to acquire a distributed lock with the specified key.
+    /// </summary>
+    /// <param name="lockKey">The lock key.</param>
+    /// <param name="expiration">The lock expiration period.</param>
+    /// <returns><see langword="true"/> if the lock was acquired; otherwise, <see langword="false"/>.</returns>
+    Task<bool> TryAcquireLockAsync(string lockKey, TimeSpan expiration);
+
+    /// <summary>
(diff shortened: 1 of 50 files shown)


axelmarquezh:
I skimmed through the changes, and they look good overall. However, I  did not run or test them.
GPT-5.5 xhigh found the following issues; take them with a grain of salt since I haven't verified them:
```
1. **Branch removed the global `IMemoryCache` registration but code still resolves it unkeyed.**
   `main` registered unkeyed `IMemoryCache`; this branch removed it. `MemoryCacheProvider` still requires `IMemoryCache`, and external tiered caches still call `ctx.Resolve<IMemoryCache>()`. This can break Autofac resolution when resolving existing string cache providers or enabling descriptor/API-client external cache.

2. **Redis degradation is incomplete: cache write failures can fail otherwise successful requests.**
   Reads in the new async external cache degrade to cache misses, but writes still throw `DistributedCacheException`. Descriptor cache misses can successfully load from the backing provider/DB and then fail while writing to Redis. API-client details external caching also still uses the synchronous provider, which throws on Redis get/set/insert failures. This behavior is partly pre-existing in `main`, but the new async/tiered path keeps the write-failure problem.

3. **`RedisDistributedLockProvider` does not recover if Lua script loading fails.**
   If startup script loading fails, `_lockReleaseScriptSha` is set to `null`. Later lock release passes that value directly to `ScriptEvaluateAsync`; despite the log saying release will retry at runtime, there is no reload path. A successfully acquired lock may fail release and remain until TTL.

4. **`RedisDistributedLockProvider` uses a plain `Dictionary` in a singleton.**
   The provider is registered as singleton and stores lock tokens in an unsynchronized `Dictionary`. Concurrent acquire/release operations can race or corrupt state.

5. **`PersonUniqueIdToUsi.BatchSize` is configurable but not validated.**
   The value is passed into Redis map cache batching and used as `i += _batchSize`. `0` causes an infinite loop; negative values can break batch allocation/copy behavior. This is branch-introduced.

6. **Person map initialization lock key is too granular.**
   The distributed initialization lock includes `MapType`, but initialization populates both map directions. Opposite resolver directions can acquire different locks and run duplicate full initializations for the same ODS/person type. A failed duplicate initializer can also reset markers after another initializer succeeded.
```

PR #1347, opened 2026-06-15
[DMS-1220] Support homonym resources and old-generator SDKs in SDK smoke tests

simpat-adam:
## Summary

Fixes two distinct defects in `EdFi.Suite3.SmokeTest.Console` (and the shared `EdFi.LoadTools` library) that block the Data Management Service full SDK smoke sweep over a Homograph-inclusive surface (DMS-1210). Both are fixed together so the resulting console build is usable by DMS.

### Fix A — HostConfiguration fallback (dual-mode factory)
The net10 upgrade (ODS-6781, `f89df4ed9`) made `SdkConfigurationFactory` require `{prefix}.Client.HostConfiguration` and **throw** when absent. New-generator SDKs ship `Client.HostConfiguration`; old-generator SDKs (the DMS `EdFi.DmsApi.TestSdk` and the published, un-regenerable `EdFi.OdsApi.Sdk`) ship only `Client.Configuration`.

- `SdkConfigurationFactory` now **probes** for `Client.HostConfiguration` first and selects the mode by which type is present — it does not rely on catching exceptions.
- When absent, it falls back to the pre-net10 path: construct `Client.Configuration` with its 4-arg ctor and set the bearer token on `AccessToken`. (The new-gen DI path is preserved verbatim.)
- `EdFiConstants` re-adds `SdkConfigurationClassNameLegacy = "Client.Configuration"`.
- `BaseTest.GetApiInstance()` now short-circuits in legacy mode to `new XxxApi(configuration)`; otherwise the existing "manual construction" block would pick the parameterless constructor and yield a token-less / wrong-base-path instance.

### Fix B — SdkCategorizer supports multi-resource API classes
openapi-generator names an API class from the OpenAPI tag, and a homonym extension puts its homonym resource on the core resource's tag — so a single `ContactsApi` carries both `Contact` and `HomographContact`.

- `SdkCategorizer.ResourceApis` now emits **one `IResourceApi` per resource** (keyed on the POST model type) instead of one per API type; `ResourceApi` resolves each verb within its resource's method group and the multiple-POST throw is removed.
- The previous `!Name.Contains("_")` filter (which silently discarded the `_N` homonym methods) is removed and a `WithHttpInfo` filter added.
- GET/DELETE — which carry no model type in their signatures — are routed by the generator's `_N` suffix plus the resource name-stem (longest stem-prefix wins). This handles **both** the `_N`-suffix (ODS) and distinct-stem (DMS) homonym conventions and bridges the singular-POST/plural-GET name mismatch.
- Single-resource API types take a fast path that preserves the original behavior exactly.

The `IResourceApi` / `ISdkCategorizer` contracts are unchanged, so all consumers (`ModelDependencySort`, `DestructiveMethodsGenerator`, `GetMethodsGenerator`, and the per-verb tests) keep their single-resource view.

## Testing

- `EdFi.SmokeTest.Console` + `EdFi.LoadTools` build clean (`TreatWarningsAsErrors`, 0 warnings).
- `EdFi.LoadTools.Test` SmokeTests: **35/35 pass**, including 5 new tests:
  - Per-resource split against the real `ContactsApi` homonym — two resources (`EdFiContact` / `HomographContact`), each verb resolving to its own `_0` / non-`_0` method, no throw.
  - Single-resource regression (`AcademicWeeksApi`).
  - `ResourceApis.ToDictionary(ModelType.Name)` regression (DestructiveMethodsGenerator / ModelDependencySort).
  - Legacy `GetApiInstance` short-circuit constructs `new XxxApi(configuration)`.

## Downstream (not in this PR; tracked under DMS-1210)
Publish a new `EdFi.Suite3.SmokeTest.Console` build to the EdFi feed, then bump the `7.3.10008` pin in the DMS `eng/smoke_test/Invoke-NonDestructiveSdkTests.ps1` and `Invoke-DestructiveSdkTests.ps1` and re-run the scheduled smoke test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -81,7 +81,7 @@
 
     <!-- Serialization -->
     <PackageVersion Include="JsonSubTypes" Version="2.0.1" />
-    <PackageVersion Include="MessagePack" Version="3.1.4" />
+    <PackageVersion Include="MessagePack" Version="3.1.7" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.4" />
 
     <!-- Logging -->
--- a/Utilities/DataLoading/EdFi.LoadTools.Test/EdFi.LoadTools.Test.csproj
+++ b/Utilities/DataLoading/EdFi.LoadTools.Test/EdFi.LoadTools.Test.csproj
@@ -20,6 +20,8 @@
     </ItemGroup>
     <ItemGroup>
         <ProjectReference Include="..\EdFi.LoadTools\EdFi.LoadTools.csproj" />
+        <!-- Referenced so the legacy branch of SdkConfigurationFactory can be tested directly. -->
+        <ProjectReference Include="..\EdFi.SmokeTest.Console\EdFi.SmokeTest.Console.csproj" />
     </ItemGroup>
   <ItemGroup Condition=" '$(StandardVersion)' == '6.1.0' ">
 		<PackageReference Include="EdFi.Suite3.OdsApi.TestSdk.Standard.6.1.0" />
--- a/Utilities/DataLoading/EdFi.LoadTools.Test/SmokeTests/BaseTestGetApiInstanceTests.cs
+++ b/Utilities/DataLoading/EdFi.LoadTools.Test/SmokeTests/BaseTestGetApiInstanceTests.cs
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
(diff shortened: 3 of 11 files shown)


Directory.Packages.props
```diff
@@ -81,7 +81,7 @@
 
     <!-- Serialization -->
     <PackageVersion Include="JsonSubTypes" Version="2.0.1" />
-    <PackageVersion Include="MessagePack" Version="3.1.4" />
+    <PackageVersion Include="MessagePack" Version="3.1.7" />
```
simpat-adam, line 84: There is a vulnerability in `3.1.4`

simpat-jesus (approved):

PR #1349, opened 2026-06-18
Redis improvements v2: explicit External/Hybrid cache mode selection

adiaz-msdf:
## Summary

Implements the FR-CACHE caching-mode requirements from the Redis Improvements PRD. The ODS/API can now independently  select the caching strategy for each of the three cache types — **Descriptors**,  **PersonUniqueIdToUsi**, and *ApiClientDetails** — between:

  - **In-memory** — `UseExternalCache: false` (unchanged default behavior)
  - **External** — L2 (Redis) only, no in-process tier
  - **Hybrid** — short-lived in-process L1 (`IMemoryCache` / `InMemoryMapCache`) in front of L2 Redis
 

    ## Changes

  **`Ed-Fi-ODS` (this PR, `redis-improvements-v2` → `redis-improvements`)**
  - `CacheSettings.cs` — new `CachingMode { External, Hybrid }` enum; per-type `CachingMode` property with behavior-preserving defaults; `L1CacheDurationSeconds` added to `PersonUniqueIdToUsiCacheConfiguration`.
  - `TieredMapCache.cs` *(new)* — L1+L2 map cache for the Person/UniqueId Redis HASH. Write-through to L2 only (bulk init loads can be millions of rows); L1 populates lazily on read and backfills only entries actually found in L2; deletes evict both.
  - `ExternalCacheModule.cs` — all three `Override*Cache` methods now branch on `CachingMode`. External-mode Descriptors register one `AsyncExternalCacheProvider` instance as both the sync and async provider so the interceptor's reference-equality check avoids a blocking Redis call on the hot path.
  - `AsyncExternalCacheProvider`, `ExternalCacheProvider`, `TieredCacheProvider`, `TieredMapCache` — Debug-level `[Hybrid]`/`[External]` L1/L2 traffic logging (guarded by `IsDebugEnabled`); plus a startup INFO summary of each type's effective mode.
  - `DistributedCacheAvailability.cs` — availability now classified by exception **type** only (dropped the brittle message-substring matching) so a genuine app bug isn't silently swallowed as a transient cache miss.
  - Tests: `TieredMapCacheTests.cs` *(new)*; extended `ExternalCacheProviderTests`.


Application/EdFi.Ods.Features/ExternalCache/TieredCacheProvider{TKey}.cs
```diff
@@ -65,15 +67,31 @@ public bool TryGetCachedObject(TKey key, out object value)
     {
         if (TryGetLocalCacheValue(key, out value))
         {
+            if (_logger.IsDebugEnabled)
+            {
+                _logger.Debug($"[Hybrid] L1 (in-memory) cache hit for key '{key}'.");
+            }
```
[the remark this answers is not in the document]
stephenfuqua: This is worth spending more time thinking about. My initial reaction is that the code should _NOT_ be altered: the debug statement needs the cache key _including the bearer token_ in order to be useful.

But we should ask for feedback from end-users. @stevenarnold-ed-fi can you check with MiDataHub and TN to get their opinions?

Application/EdFi.Ods.Features/ExternalCache/AsyncExternalCacheProvider{TKey}.cs
```diff
@@ -97,13 +122,70 @@ await _distributedCache.SetStringAsync(
                     })
                 .ConfigureAwait(false);
         }
+        catch (Exception ex) when (DistributedCacheAvailability.IsUnavailable(ex))
+        {
+            _logger.Warn($"Distributed cache unavailable; skipping cache write for key '{keyAsString}'.", ex);
+        }
```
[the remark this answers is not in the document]
stephenfuqua: Now this is different: a WARNING should not risk including the bearer token. In fact, I wonder if we actually need this much detail in the warning log? What diagnostic value comes from knowing exactly which cache write missed? There is no mechanism to manually retry. I suspect the information is irrelevant at this point. But again, good to hear from actual system administrators before deciding.

Application/EdFi.Ods.Features/ExternalCache/TieredCacheProvider{TKey}.cs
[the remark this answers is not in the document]
stevenarnold-ed-fi: Yep, Emails sent to both!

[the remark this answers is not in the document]
adiaz-msdf: Added CacheKeyLogSanitizer.Redact(key): renders ApiClientDetails keys as "ApiClientDetails.<redacted>", otherwise return the key id.

Application/EdFi.Ods.Features/ExternalCache/AsyncExternalCacheProvider{TKey}.cs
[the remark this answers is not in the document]
adiaz-msdf: Removed the key in the Warn messages, just the generic text + exception. The correlationId can comes from the log4net layout (%property{CorrelationId}), which is the tracking mechanism see https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pull/1392

Application/EdFi.Ods.Features/ExternalCache/AsyncExternalCacheProvider{TKey}.cs
```diff
@@ -67,14 +82,22 @@ public class AsyncExternalCacheProvider<TKey>(
...
                     ExternalCacheSerializationHelper.Serialize(obj),
                     CreateDistributedCacheEntryOptions())
                 .ConfigureAwait(false);
         }
+        catch (Exception ex) when (DistributedCacheAvailability.IsUnavailable(ex))
+        {
+            // A failed cache write must not fail the request — skip caching when the cache is
+            // temporarily unavailable (open circuit breaker or Redis connectivity failure).
+            _logger.Warn($"Distributed cache unavailable; skipping cache write for key '{keyAsString}'.", ex);
+        }
```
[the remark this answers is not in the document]
adiaz-msdf: fixed, removed the key on warn exceptions

Application/EdFi.Ods.Common/Configuration/CacheSettings.cs
```diff
@@ -72,4 +101,20 @@ public enum ExternalCacheProviderOption
...
+    {
+        /// <summary>
+        /// Use the external (L2) cache directly, with no in-process L1 tier.
+        /// </summary>
+        External,
+
+        /// <summary>
+        /// Front the external (L2) cache with a short-lived in-process L1 cache.
+        /// </summary>
+        Hybrid,
```
stephenfuqua, line 150: Should also have "InMemory" for L1 only.

Oh, and should it also have a "None" option if you don't want any caching at all?

Application/EdFi.Ods.Features/ExternalCache/CacheKeyLogSanitizer.cs
```diff
@@ -0,0 +1,33 @@
...
+internal static class CacheKeyLogSanitizer
+{
+    /// <summary>
+    /// Returns a log-safe rendering of a cache key, redacting the API client credential embedded in
+    /// ApiClientDetails cache keys (<c>ApiClientDetails.&lt;apiClientKey&gt;</c>) so it never reaches the logs.
+    /// Non-string keys (e.g. descriptor <see cref="ulong"/> ids and Person map-cache tuple keys) carry no
+    /// secret and are rendered with <see cref="object.ToString"/>.
+    /// </summary>
+    /// <param name="key">The cache key to render.</param>
+    public static string Redact(object key)
```
stephenfuqua: The function name `Redact` made me think that _all_ keys would be redacted - but it is really only for client tokens.

Suggestion: rename to `RedactClientTokenFromKey`, or to something more generic like `Sanitize`.

Application/EdFi.Ods.Features/ExternalCache/AsyncExternalCacheProvider{TKey}.cs
```diff
@@ -97,13 +122,70 @@ await _distributedCache.SetStringAsync(
...
     }
 
+    // Runs the distributed-cache read through the Redis circuit breaker when resilience is configured,
+    // so repeated failures fail fast once the circuit opens instead of blocking on the full timeout.
+    private async Task<string> ExecuteGetAsync(string keyAsString)
+    {
+        if (_resilience is null)
+        {
+            return await _distributedCache.GetStringAsync(keyAsString).ConfigureAwait(false);
+        }
```
stephenfuqua: I'm curious - why provide the option for resilience provider to be null? Wouldn't we _always_ want that? This gives us an extra code path to test, which might be unnecessary.

Application/EdFi.Ods.Features/ExternalCache/CacheKeyLogSanitizer.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if (key is null)
+        {
+            return "<null>";
+        }
```
stephenfuqua, line 41: Is `null` possible here? I'm a little rusty with my C#, but I would have though this is impossible since the function signature is not `Redact(object? key)`.

Application/EdFi.Ods.Features/ExternalCache/CacheKeyLogSanitizer.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        return key is string s && s.StartsWith(ApiClientDetailsCacheKeyProvider.CacheKeyPrefix, StringComparison.Ordinal)
+            ? ApiClientDetailsCacheKeyProvider.CacheKeyPrefix + "<redacted>"
```
stephenfuqua: An end-user, in another setting, asked if we could "log the last four to six digits of the token."

I think that is a safe thing to do. Instead of replacing with "<redacted>", I would use "xxx" to help indicate that something was deleted, then keep the last six characters.

This token might be a JSON Web Token now, which is quite long. There wouldn't be value to replacing every single character with `x`. But putting a few `x` characters in there helps the reader know that there previously something else there, which we are now hiding.

Application/EdFi.Ods.Features/ExternalCache/CacheKeyLogSanitizer.cs
adiaz-msdf: Yes, objects are nullable by default in C#, even when not marked as optional

Application/EdFi.Ods.Common/Configuration/CacheSettings.cs
adiaz-msdf: I didn’t add InMemory/None because I kept the UseExternalCache setting for backward compatibility. Essentially, InMemory/None is equivalent to UseExternalCache: false. But if you’d prefer, I can add the “None” option to make it more explicit. @stephenfuqua

Application/EdFi.Ods.Features/ExternalCache/AsyncExternalCacheProvider{TKey}.cs
adiaz-msdf: The optional parameter only exists to simplify construction in the two places that have no resilience pipeline: the unit tests, and the lightweight `**TieredCacheProvider** `convenience constructor. In the actual DI registrations (**ExternalCacheModule**, both the `External :178` and `Hybrid :159` paths) we always resolve and pass **RedisCacheResilience**, so production always runs through the circuit breaker. The `_resilience is null `branch is just a direct pass-through for those non-production cases. I kept it optional to avoid forcing every test to stand up a Polly pipeline — but if you'd rather have a single code path I can make it required and build a default pipeline in the tests/convenience ctor. Please confirm if we need this change @stephenfuqua

stephenfuqua: Suggestion: extract an interface for `RedisCacheResilience` and then you can use FakeItEasy to mock the object, instead of using Polly in the unit tests.

Application/EdFi.Ods.Common/Configuration/CacheSettings.cs
stephenfuqua: Nice to have: `"none"` as an explicit text, but treat `""` (empty string) as equivalent of `"none"` so that existing users with an empty string do not have to change their appsettings.

It helps for me to look at the [other pull request](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Implementation/pull/1387/changes#diff-b17c2ad157cd289159bae3b510bfb199bd0be3f1760c9655a93228bd1700ad2f) at the same time to see the `appsettings.json` file. This reminds me that the value is going into a key called `ExternalCacheProvider`, where it would not make sense to have `"InMemory"` as an option. 

Conclusion: yes, I agree with your analysis. Nice-to-have addition of `"none"` as synonym for `""`.

Application/EdFi.Ods.Features/ExternalCache/CacheKeyLogSanitizer.cs
adiaz-msdf: fixed

Application/EdFi.Ods.Features/ExternalCache/CacheKeyLogSanitizer.cs
adiaz-msdf: fixed

Application/EdFi.Ods.Common/Configuration/CacheSettings.cs
adiaz-msdf: Fixed: Added None for CachingMode An empty or unspecified value is treated as None

Application/EdFi.Ods.Features/ExternalCache/AsyncExternalCacheProvider{TKey}.cs
adiaz-msdf: Fixed: Made `RedisCacheResilience `required in the `ExternalCacheProvider`, removed unnesesary code path to test

